PR #1, opened 2016-08-30
rename gem to represent loading path

durandom:
with the name `xclarity` there has to be a `lib/xclarity.rb` to make the gem load / init.

this naming here follows the miq naming scheme


juliancheal:
LGTM 👍

PR #2, opened 2016-09-12
Provider modeling

terciodemelo:
Hello there, we've being doing some advances on implementing the provider it's not complete yet but we're actively working on it. 

The following aspects have been implemented:
* The provider is being successfully saved on manage iq
* Some model tests are implemented like `ems_type`, `description`, and `connect` testing. On the other hand, some are missing such as of authentication verification and discovery.
* Controller test from ManageIQ are integrated into the gem

It is not clear to us if we should consider LXCA as an InfraManager, or being a PhysicalInfraManager, it should be in a completely separated group from all the other existent providers for MIQ. For now the related parts of our contributions will be commited to this PR. Others such as the refresher in different branches.


juliancheal:
> It is not clear to us if we should consider LXCA as an InfraManager, or being a PhysicalInfraManager

While we're creating LXCA I'm simultaneously adding support in ManageIQ for PhysicalInfraManagers. https://github.com/ManageIQ/manageiq/pull/10807

juliancheal:
Would you be able to squash some of these commits, for instance I think the deletions in https://github.com/juliancheal/manageiq-providers-lenovo/pull/2/commits/4f8b78f2cb115fae60ea87424b473e60779702e7 get added back in.

And https://github.com/juliancheal/manageiq-providers-lenovo/pull/2/commits/25a289a2505d9ed5af3f417025cb96309ed60187 could be squashed into https://github.com/juliancheal/manageiq-providers-lenovo/pull/2/commits/740e4bbebd6798735d1827560650b4117fe5cce0

terciodemelo:
Sure enough.

terciodemelo:
I've made some cleaning up in other commits as well

PR #5, opened 2016-10-12
Add Services

Fryguy:
Services have been enabled, so this PR adds the badges and enables Travis.

@chessbyte Please review.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,12 @@
+language: ruby
+rvm:
+- '2.3.1'
+sudo: false
+cache: bundler
+notifications:
+  webhooks:
+    urls:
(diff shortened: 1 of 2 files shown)


Fryguy:
Even though Travis is failing, it might be better to merge and let master go red, and then the first thing after that would be to get master to go green.  cc @juliancheal @durandom

juliancheal:
So the reason it's failing is due to the dependency on xclarity_client. My plan today is to push that up to Rubygems as a 0.1 version gem. As we still have lots more discussions/code to do on the gem, so waiting until then would be too long.

juliancheal:
My #6 PR should fix the issues with this PR.

PR #15, opened 2016-11-02
Updated XClarity gem to the lastest version

juliancheal:


app/models/manageiq/providers/lenovo/manager_mixin.rb
```diff
@@ -11,11 +11,11 @@ def description
...
 
-    username = options[:user] || authentication_userid(options[:auth_type])
-    password = options[:pass] || authentication_password(options[:auth_type])
-    host     = options[:host]
-
-    self.class.raw_connect(username, password, host)
+    username   = options[:user] || authentication_userid(options[:auth_type])
+    password   = options[:pass] || authentication_password(options[:auth_type])
+    host       = options[:host]
+    verify_ssl = options[:verify_ssl] ||
```
Fryguy: Rubocop: Useless assignment to variable - `verify_ssl`

app/models/manageiq/providers/lenovo/manager_mixin.rb
```diff
@@ -11,11 +11,11 @@ def description
...
+    self.class.raw_connect(username, password, host, verify_ssl)
```
Fryguy: And this one is part of the `||` on the previous line so should be indented  (unless that `||` is the actual bug here)

app/models/manageiq/providers/lenovo/manager_mixin.rb
```diff
@@ -52,7 +53,7 @@ def discover(username, password, host)
       all_emses         = includes(:authentications)
       all_ems_names     = all_emses.map(&:name).to_set
 
-      xclarity = raw_connect(username, password, host)
+      xclarity = raw_connect(username, password, host, verify_ssl)
```
Fryguy: Rubocop: Useless assignment to variable - `xclarity`.

app/models/manageiq/providers/lenovo/manager_mixin.rb
juliancheal: Ha I just went to see why this was there, and realised the file isn't checked in with the extra `||`

PR #25, opened 2017-01-20
Ran provider (update) generator.

juliancheal:
This keeps the provider inline with new updates for all providers.
Also removed rubocop_base.yml as no longer required.

PR #26, opened 2017-01-20
Set correct base class for Lenovo Manager

blomquisg:
The base manager class for `Lenovo::PhysicalInfraManager` was accidentally left as `InfraManager`.  This was a hold over from before the base `PhysicalInfraManager` class existed.

By updating to the correct base class, functions like `ManageIQ::Providers::PhysicalInfraManager.supported_subclasses` and `ManageIQ::Providers::PhysicalInfraManager.supported_types_and_descriptions_hash` will now return the correct list.  For example:

```
irb(main):003:0> ManageIQ::Providers::PhysicalInfraManager.supported_subclasses.map(&:name)
=> ["ManageIQ::Providers::Lenovo::PhysicalInfraManager"]
irb(main):004:0> ManageIQ::Providers::PhysicalInfraManager.supported_types_and_descriptions_hash
=> {"lenovo_ph_infra"=>"Lenovo XClarity"}
```


juliancheal:
LGTM 👍 and all that jazz. Travis is Red, but that's because I need to merge https://github.com/ManageIQ/manageiq-providers-lenovo/pull/25. The failure is because a file was moved to ui-classic from manageiq repo.

PR #30, opened 2017-02-23
Adds event collection to lenovo provider

rodneyhbrown7:
@miq-bot add_label: wip, providers/physical-infrastructure

### DIFF
--- a/app/models/manageiq/providers/lenovo/physical_infra_manager/event_catcher/stream.rb
+++ b/app/models/manageiq/providers/lenovo/physical_infra_manager/event_catcher/stream.rb
@@ -1,5 +1,4 @@
 class ManageIQ::Providers::Lenovo::PhysicalInfraManager::EventCatcher::Stream
-
   # Creates an event monitor
   #
   def initialize(ems)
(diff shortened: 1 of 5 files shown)


juliancheal:
@miq-bot add_label wip, providers/physical-infrastructure

app/models/manageiq/providers/lenovo/physical_infra_manager/event_catcher/stream.rb
```diff
@@ -0,0 +1,51 @@
...
+    yield get_events.collect { |e| ManageIQ::Providers::Lenovo::PhysicalInfraManager::EventParser.event_to_hash(e, @ems.id) }
+  end
+
+  def event_monitor_handle
+    @event_monitor_handle ||= create_event_monitor_handle @ems
+  end
+
+  private
+
+  def get_events
```
juliancheal: We should rename this to `events`

juliancheal:
This LGTM apart from renaming `get_event` to `events`. There was also some discussion over on the ManageIQ channel about using the Lonely Operator `&.` https://gitter.im/ManageIQ/manageiq?at=58b99fb021d548df2c8797b2

I think the use of it here is fine, but I'd like a second opinion @chrisarcand.

chrisarcand:
The use of it here is correct.

PR #34, opened 2017-03-17
refactoring of refresh

juliancheal:
Added Firmware model

Started to extract attributes to add to other tables such as:
hardwares, networks, etc.

PR #37, opened 2017-03-24
Associate ph server to host

walteraa:
Now the relationship is associated in Provider.

The relationship will be assign only if the physical server exists previously in Database and if exists a host . So, the relationship will be done in the second refresh for each Physical server.

This PR depends of https://github.com/ManageIQ/manageiq-providers-lenovo/pull/34 and maybe some change should be done after merge.


walteraa:
Reopen to rebuild.

walteraa:
Reopen to rebuild.

walteraa:
@juliancheal @rodneyhbrown7 I am having migrations erros when I try to execute the refresh process. Is there any PR referring to migrations that solve these issues? Is missing only test if the refresh process will assign the physical server ID in host.

rodneyhbrown7:
Walter what errors are you seeing in the refresh process?

walteraa:
@rodneyhbrown7 

`EmsRefresh::Refreshers::EmsRefresherMixin::PartialRefreshError: Cannot modify association 'ManageIQ::Providers::Lenovo::PhysicalInfraManager::PhysicalServer#hardware' because the source reflection class 'Hardware' is associated to 'ComputerSystem' via :has_one.`

rodneyhbrown7:
Do you have the PR  14430 installed?

rodneyhbrown7:
https://github.com/ManageIQ/manageiq/pull/14430  and https://github.com/ManageIQ/manageiq-providers-lenovo/pull/34

walteraa:
@rodneyhbrown7 I'm running MIQ core from `julian/physical_provider_refresh bce86f9 "Fixed refresh for Physical Infra"` and My provider branch was merged with  `juliancheal:refactor_refresh` in your last commit.

app/models/manageiq/providers/lenovo/physical_infra_manager/refresh_parser.rb
```diff
@@ -27,9 +28,79 @@ def ems_inv_to_hashes
...
+      }
+      new_result[:hardware] = get_hardwares(node)
+      save_host_relationship(node)
+      return node.uuid, new_result
+    end
+
+    def save_host_relationship(node)
+      # Assign a physicalserver and host if server already exists and
+      # some host match with physical Server's serial number
+      server = PhysicalServer.find_by(:uuid => node.uuid, :ems_id => @ems.id)
```
rodneyhbrown7: :uuid is not part of the PhysicalServer definition. I think this needs to be changes to  find_by(:ems_ref => node.uuid

rodneyhbrown7: A physical server is associated to a ComputerSystem which has a hardware component.  I don't think that this assignment here is valid. I think we need to assign the hardware to the computer system and then the computer system to the physical server. See L80

app/models/manageiq/providers/lenovo/physical_infra_manager/refresh_parser.rb
```diff
@@ -27,9 +28,79 @@ def ems_inv_to_hashes
...
+        :product_name           => node.productName,
+        :manufacturer           => node.manufacturer,
+        :machine_type           => node.machineType,
+        :model                  => node.model,
+        :serial_number          => node.serialNumber,
+        :field_replaceable_unit => node.FRU,
+        # Filled in later conditionally on what's available
+        :computer_system        => {:hardware => {:networks => [], :firmwares => []}}
+      }
+      new_result[:hardware] = get_hardwares(node)
```
rodneyhbrown7: The hardware element is not associated to the physical server directly.  Hardware is associated to ComputerSystem which is associated to the PhysicalServer (see L80) .   Rails will not resolve this assignment.

app/models/manageiq/providers/lenovo/physical_infra_manager/refresh_parser.rb
walteraa: Is the `:uuid` removed from Physical server?

app/models/manageiq/providers/lenovo/physical_infra_manager/refresh_parser.rb
walteraa: This assignment come from @juliancheal's branch. What do you think about @rodneyhbrown7's concern, Julian?

rodneyhbrown7: If you look at  https://github.com/ManageIQ/manageiq-providers-lenovo/pull/34    This assignment is now new_result[:computer_system][:hardware] = get_hardwares(node).  It may be worth rebasing this PR from 34 before it is merged.

walteraa: Ok, I'll do a rebase, @rodneyhbrown7, thanks!

walteraa: Done.

app/models/manageiq/providers/lenovo/physical_infra_manager/refresh_parser.rb
walteraa: I changed it to `:ems_ref`.

app/models/manageiq/providers/lenovo/physical_infra_manager/refresh_parser.rb
```diff
@@ -76,17 +76,28 @@ def parse_physical_server(node)
...
         :serial_number          => node.serialNumber,
         :field_replaceable_unit => node.FRU,
-        :computer_system        => {
-          :hardware             => {
-            :networks  => [],
-            :firmwares => [] # Filled in later conditionally on what's available
-          }
-        }
+        # Filled in later conditionally on what's available
+        :computer_system        => {:hardware => {:networks => [], :firmwares => []}}
```
rodneyhbrown7: This change will be functional but removes formatting that was added previously. Rubocop will probably complain about laying this out on a single line like this.

walteraa: @rodneyhbrown7 I did this change because rubocop was warning about the previous structure made by Julian. Rubocop doesn't warn about this format.

rodneyhbrown7: Ok. Good. Thanks Walter.

app/models/manageiq/providers/lenovo/physical_infra_manager/refresh_parser.rb
```diff
@@ -76,17 +76,28 @@ def parse_physical_server(node)
...
       }
       new_result[:computer_system][:hardware] = get_hardwares(node)
+      save_host_relationship(node)
       return node.uuid, new_result
     end
 
+    def save_host_relationship(node)
```
juliancheal: I believe we should just be finding hosts here, but not saving them. All saving should be done in ManageIQ in `SaveInventory`.

walteraa: @juliancheal I don't figured out how to send a save host to MIQ core from here. I mean, the ID of relationship was saved in Physical server, as you guys defined, so, once we found the Host ID and the Physical server ID which matches considering the LXCA association(serial_number and service_tag equals), how to do that in Core?

rodneyhbrown7: I think you do the look up and assignments in the refresher as you have already implemented, but you remove the line that saves the server instance.     I think that the existing code in save_ems_physical_infra_inventory (https://github.com/ManageIQ/manageiq/blob/master/app/models/ems_refresh/save_inventory_physical_infra.rb#L8 )  should save the details without any additional changes.  This needs to be verified.

walteraa: Hey @rodneyhbrown7 , I figured out this process(which use the core to save the objects) before, it isn't the problem. The problem is that the relationship is saved in the Host object, so to use the `ems_refresh/save_inventory_physical_infra.rb` is unfeasible considering the structure proposed by RH guys.

The only way that we can use this, is converting the Host object found in a Hash, adding it as a "sub hash" in the Physical server hash and setting up a child key to host in `save_physical_server` method in `ems_refresh/save_inventory_physical_infra.rb` file.

But it looks like a workaround.

rodneyhbrown7: I don't think it is a work around. I think that is the way this is intended to work.    I believe all we need to do is remove the save! operation in the refresher and then add the :host as a child key in save_physical_server.  This is working for me locally.

walteraa: Ok @rodneyhbrown7, I'll do that as I described.

AndreyMenezes:
@juliancheal can you merge this PR? @rodneyhbrown7

juliancheal:
LGTM 👍

PR #41, opened 2017-04-07
Remove method "name" from physical_server

walteraa:
The method "name" in physical server class is causing inconsistency when we are trying show a Physical server's name in the UI and in the REST API as well.

The method was removed to solve this issue.

### DIFF
--- a/app/models/manageiq/providers/lenovo/physical_infra_manager/physical_server.rb
+++ b/app/models/manageiq/providers/lenovo/physical_infra_manager/physical_server.rb
@@ -1,7 +1,4 @@
 module ManageIQ::Providers
   class Lenovo::PhysicalInfraManager::PhysicalServer < ::PhysicalServer
-    def name
-      "physical_server"
-    end
   end
 end


juliancheal:
Hmm, I thought `name` was needed in the model for ManageIQ. Maybe @Fryguy knows why, I'm trying to find out where.

This is another reason why we need to add more tests :)

rodneyhbrown7:
walter is there a reason we can't set the name method to return the information that we want it to return?

walteraa:
The problem is that when we try to use `physical_server_object.name` if there is a method named `name` in `ManageIQ::Providers::Lenovo::PhysicalInfraManager::PhysicalServer`, this method will be called instead of the property `name` which reffering to the column `name` in the `physical_servers` table and is where the real Physical Server name value is saved, then will always return the content defined in the method, wich in this case is the string `"physical_server"`

juliancheal:
@fryguy ah ok. Must have been a mistake on my part

juliancheal:
@miq-bot add_label fine/yes

simaishi:
Fine backport details:

```
$ git log -1
commit 1fe3cb8f7659d6527d6b14c433414b5a50710b72
Author: Julian Cheal <julian.cheal@gmail.com>
Date:   Tue Apr 11 16:00:10 2017 +0100

    Merge pull request #41 from lenovo/fix_physical_server_name
    
    Remove method "name" from physical_server
    (cherry picked from commit 85538905bdbf6bc3d7c18165dfce18acf79704e9)
```

PR #43, opened 2017-04-17
Parse asset details from XClarity Client

MaysaMacedo:
This PR is able to parse the following info from XClarity Client:
- contact
- description
- location
- room 
- rack
- lowest rack unit

depends on #[14827](https://github.com/ManageIQ/manageiq/pull/14827)


app/models/manageiq/providers/lenovo/physical_infra_manager/refresh_parser.rb
```diff
@@ -124,7 +124,8 @@ def parse_physical_server(node)
             :firmwares => [] # Filled in later conditionally on what's available
           }
         },
-        :asset_details          => parse_asset_details(node)
+        :asset_details          => parse_asset_details(node),
+        :location_led_state    	=> find_loc_led_state(node.leds)
```
rodneyhbrown7: This change was merged into https://github.com/ManageIQ/manageiq-providers-lenovo/pull/48 yesterday.

MaysaMacedo: It came from this merge, I've just fixed an alignment on this line that the rubocop pointed.

juliancheal:
@MaysaMacedo is this data we already get from the current version of the `xclarity` gem or do we need to update the gem version too?

MaysaMacedo:
@juliancheal No need to update it.

PR #49, opened 2017-05-05
Physical Server model was missing from factory girl definition.

juliancheal:


juliancheal:
Ah PhysicalServer is defined in ManageIQ repo.

PR #51, opened 2017-05-08
Fix unused HTTP interactions problem

skovic:
ManageIQ now seems to be using webmock-2.3.2. In this new version, it appears that the authentication details (userid/password) are no longer included in the URI of the generated HTTP request. The exclusion of the authentication details, causes webmock to not use the requests stored in the cassettes because they now differ (the newly generated requests do not include the authentication details, but the requests stored cassettes do). In the prior version, webmock-1.24.6, the authentication details were included, so we did not have this problem.


skovic:
@miq-bot assign @juliancheal

PR #52, opened 2017-05-12
Do not define the same key twice within hash

isimluk:
It's no good.

Addressing:
`refresh_parser.rb:118: warning: key :computer_system is duplicated and overwritten on line 123`

### DIFF
--- a/app/models/manageiq/providers/lenovo/physical_infra_manager/refresh_parser.rb
+++ b/app/models/manageiq/providers/lenovo/physical_infra_manager/refresh_parser.rb
@@ -115,7 +115,6 @@ module ManageIQ::Providers::Lenovo
         :model                  => node.model,
         :serial_number          => node.serialNumber,
         :field_replaceable_unit => node.FRU,
-        :computer_system        => {:hardware => {:networks => [], :firmwares => []}},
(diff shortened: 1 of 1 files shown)


juliancheal:
I'm not sure why we're getting this error and not the other providers. As I took the idea from `openstack` where they do the same. C.F.

https://github.com/ManageIQ/manageiq-providers-openstack/blob/master/app/models/manageiq/providers/openstack/cloud_manager/refresh_parser.rb#L327 https://github.com/ManageIQ/manageiq-providers-openstack/blob/master/app/models/manageiq/providers/openstack/cloud_manager/refresh_parser.rb#L330
https://github.com/ManageIQ/manageiq-providers-openstack/blob/master/app/models/manageiq/providers/openstack/cloud_manager/refresh_parser.rb#L344

Happy to merge, but I think defining the keys first in one location looks better, but happy to be wrong :)

rodneyhbrown7:
@juliancheal  I think there was a bad merge at some point.  In our case, computer_system is defined twice within the same hash. That is what this PR would fix.  I don't think the pattern you implemented is at fault. In the prior committed this was on line 118 and 123.

PR #56, opened 2017-05-18
Fixes the definition of spec.files in gemspec

NickLaMuro:
With `Dir["{app,config.lib}"]`, that only tries to find two dirs: `app`, and `config.lib` (this one isn't real, obviously).

With the fix to `Dir["{app,config.lib}"]`, it then will search through the app, config, and lib directories properly


juliancheal:
Wonder if that typo came from the generator.

PR #63, opened 2017-07-19
run provider generator to sync changes from core

durandom:
ran the generator again with

```bash
bundle exec rails g provider <name> --path ../
```

these are the changes.

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -1,0 +1,12 @@
+---
+:ems:
+  :ems_lenovo:
+    :blacklisted_event_names: []
+    :event_handling:
+      :event_groups:
+:http_proxy:
+  :lenovo:
+    :host:
+    :password:
+    :port:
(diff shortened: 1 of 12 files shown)


config/settings.yml
```diff
@@ -0,0 +1,12 @@
...
+  :ems_lenovo:
+    :blacklisted_event_names: []
+    :event_handling:
+      :event_groups:
+:http_proxy:
+  :lenovo:
+    :host:
+    :password:
+    :port:
+    :user:
```
Fryguy: Are there any settings that have to come over from ManageIQ?

durandom: yes there are some, but should be part of a different PR

durandom: created https://github.com/ManageIQ/manageiq-providers-lenovo/issues/65

PR #69, opened 2017-07-28
Add provider support for additional power operations

skovic:
Update provider and spec tests to support additional power operations.


skovic:
@miq-bot assign @juliancheal

PR #82, opened 2017-09-21
Add event_id to the event parser and change the query for refresh eve…

AndreyMenezes:
This PR do:
- Add ems_ref to the event_parser and assign value of `cn`. This attribute is the unique sequential ID for the event for each LXCA. So the `ems_id + ems_ref` are unique for each event in the table `event_streams`.
captura de tela 2017-09-22 as 15 44 10

- Change the query for the refresh events to filter using the ems_ref.

Depends on:
https://github.com/ManageIQ/manageiq-schema/pull/70

This image show why we need to use `cn` attribute to differentiate between events with the same type.
The event from left of the print screen was added to `event_streams` and the right side event not, this occurs because of each identify until now for the event in the MIQ could be the same (ems_id, timestamp, event_type...)
pasted image at 2017_09_22 03_39 pm

### DIFF
--- a/app/models/manageiq/providers/lenovo/physical_infra_manager/event_catcher/stream.rb
+++ b/app/models/manageiq/providers/lenovo/physical_infra_manager/event_catcher/stream.rb
@@ -55,6 +55,6 @@ class ManageIQ::Providers::Lenovo::PhysicalInfraManager::EventCatcher::Stream
   end
 
   def get_last_cnn_from_events(ems_id)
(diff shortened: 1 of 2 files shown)


CharlleDaniel:
@AndreyMenezes morning, cloud you rebase your branch with the master, please?

AndreyMenezes:
@juliancheal @blomquisg The dependencies were resolved :)

juliancheal:
LGTM 👍

PR #84, opened 2017-09-28
[WIP] Update the uid_ems to the uuid of the endpoint

rodneyhbrown7:
Fills in the uuid of the endpoint with a UUID that is unique to this instance of the provider. Currently this value is empty. To be more correct this field should be populated.

@agrare , @AndreyMenezes

### DIFF
--- a/app/models/manageiq/providers/lenovo/physical_infra_manager/refresh_parser.rb
+++ b/app/models/manageiq/providers/lenovo/physical_infra_manager/refresh_parser.rb
@@ -272,7 +272,7 @@ module ManageIQ::Providers::Lenovo
         :type                   => "ManageIQ::Providers::Lenovo::PhysicalInfraManager::PhysicalServer",
         :name                   => node.name,
         :ems_ref                => node.uuid,
-        :uid_ems                => @ems.uid_ems,
+        :uid_ems                => node.uuid,
         :hostname               => node.hostname,
         :product_name           => node.productName,
         :manufacturer           => node.manufacturer,


rodneyhbrown7:
@miq-bot add_label wip

app/models/manageiq/providers/lenovo/physical_infra_manager/refresh_parser.rb
```diff
@@ -272,7 +272,7 @@ def parse_physical_server(node)
         :type                   => "ManageIQ::Providers::Lenovo::PhysicalInfraManager::PhysicalServer",
         :name                   => node.name,
         :ems_ref                => node.uuid,
-        :uid_ems                => @ems.uid_ems,
+        :uid_ems                => node.uuid,
```
AndreyMenezes: We shouldn't have two variables with the same value. `uid_ems` was never used, I believe we can remove this.

What's `@ems.uid_ems` represents @juliancheal ?

walteraa: @AndreyMenezes  according to the file `models/manager_refresh/inventory/core.rb` at line `62`, and some converstations that I had with some RH guys, the `uid_ems` is the reference of the manager in the database.

walteraa:
@rodneyhbrown7 I think this PR should be closed. `uid_ems` **shouldn't** references `node.uuid` because `uid_ems` is the way to join the inventory objects and the Manager.

app/models/manageiq/providers/lenovo/physical_infra_manager/refresh_parser.rb
agrare: It depends on the model, the idea is that ems_ref is how the provider references the object and uid_ems is something we can use to reference the object across providers.  It is okay if they are the same value as long as that value fills both roles.

The reference to the manager is ems_id.

walteraa: Got it @agrare it was the guys passed to me. In this case, I think it's LGTM.

walteraa:
LGTM

app/models/manageiq/providers/lenovo/physical_infra_manager/refresh_parser.rb
AndreyMenezes: @agrare If it's not necessary `uid_ems` to refer the provider, in this case I believe they could be the same value.

rodneyhbrown7:
Thanks for the comments.  RH team did comment during our interlock that this value should be filled in and we agreed that the UUID would be decent value to use.   @AndreyMenezes can you approve the merge request to complete this story?

PR #87, opened 2017-10-10
Update the refresh_parser to support the retrieval of config patterns

skovic:
This PR updates the refresh_parser, so it can retrieve config patterns from a LXCA instance.


skovic:
@AndreyMenezes Thanks for taking a look. That acronym for the local variable in the test just stands for "refresh parser". It's being used in some of the other test cases in that file.

AndreyMenezes:
@skovic Right, I just don't like "unknown" acronyms. Internally we can understand, but It's not clear to the others people.

rodneyhbrown7:
LGTM

PR #89, opened 2017-10-13
Test the end-to-end refresh in the refresher_spec

agrare:
Instead of just testing parse with vcr, then saving no inventory we can
run the full EmsRefresh.refresh in the vcr block so we can test
relationship links created in save_ems_inventory in the future.


agrare:
Hm this is passing locally, not sure why failing on travis

rodneyhbrown7:
@agrare  Can you close and reopen this PR to see if it will pass travis tests?

agrare:
Same problem, wondering if there are some difference between how the parser was invoked by the original test vs a full refresh.

rodneyhbrown7:
I believe this is failing because there is a request in the cassette that is unused. See:
https://github.com/agrare/manageiq-providers-lenovo/blob/6efc54050f4f9330ac0562f8995840cdaf15cd88/spec/vcr_cassettes/manageiq/providers/lenovo/physical_infra_manager/refresher.yml#L290

If you delete the second request I expect this test to pass.

agrare:
Closing now that https://github.com/ManageIQ/manageiq-providers-lenovo/pull/138 does a full refresh test

PR #93, opened 2017-10-26
searches the hardware resource for a relationship

rodneyhbrown7:
To address the bug:  https://bugzilla.redhat.com/show_bug.cgi?id=1505590

Host relationships are not created to the RHEV VM instances because the service_tag field in the host table is not set. This value is set in the hardwares table associated with the host.   Joining the two tables to find the hardwares.serial_number field that matches the physical server this host is sitting on.

@miq-bot add_label wip

### DIFF
--- a/app/models/manageiq/providers/lenovo/physical_infra_manager/refresh_parser.rb
+++ b/app/models/manageiq/providers/lenovo/physical_infra_manager/refresh_parser.rb
@@ -58,7 +58,119 @@ module ManageIQ::Providers::Lenovo
 
     def get_config_patterns
       config_patterns = @connection.discover_config_pattern
-      process_collection(config_patterns, :customization_scripts) { |config_pattern| @parser.parse_config_pattern(config_pattern) }
+      process_collection(config_patterns, :customization_scripts) { |config_pattern| parse_config_pattern(config_pattern) }
+    end
+
+    def parse_firmware(firmware)
+      {
+        :name         => "#{firmware["role"]} #{firmware["name"]}-#{firmware["status"]}",
+        :build        => firmware["build"],
+        :version      => firmware["version"],
+        :release_date => firmware["date"],
+      }
+    end
+
+    def parse_management_device(node)
+      {
+        :device_type => "management",
+        :network     => parse_management_network(node),
+        :address     => node.macAddress
+      }
+    end
+
+    def parse_management_network(node)
+      {
+        :ipaddress   => node.mgmtProcIPaddress,
+        :ipv6address => node.ipv6Addresses.nil? ? node.ipv6Addresses : node.ipv6Addresses.join(", ")
+      }
+    end
+
+    def parse_addin_cards(addin_card)
+      {
+        :device_name            => addin_card["productName"],
+        :device_type            => get_device_type(addin_card),
+        :firmwares              => get_guest_device_firmware(addin_card),
+        :manufacturer           => addin_card["manufacturer"],
+        :field_replaceable_unit => addin_card["FRU"],
+        :location               => "Bay #{addin_card['slotNumber']}",
+        :child_devices          => get_guest_device_ports(addin_card)
+      }
+    end
+
+    def parse_onboard_devices(onboard_device)
+      {
+        :device_name   => onboard_device["name"],
+        :device_type   => get_device_type(onboard_device),
+        :firmwares     => get_guest_device_firmware(onboard_device),
+        :location      => "Onboard",
+        :child_devices => get_guest_device_ports(onboard_device)
+      }
+    end
+
+    def parse_physical_port(port)
+      {
+        :device_type => "physical_port",
+        :device_name => "Physical Port #{port['physicalPortIndex']}"
+      }
+    end
+
+    def parse_logical_port(port)
+      {
+        :address => format_mac_address(port["addresses"])
+      }
+    end
+
+    def format_mac_address(mac_address)
+      mac_address.scan(/\w{2}/).join(":")
+    end
+
+    def parse_physical_server(node)
+      new_result = {
+        :type                   => "ManageIQ::Providers::Lenovo::PhysicalInfraManager::PhysicalServer",
+        :name                   => node.name,
+        :ems_ref                => node.uuid,
+        :uid_ems                => node.uuid,
+        :hostname               => node.hostname,
+        :product_name           => node.productName,
+        :manufacturer           => node.manufacturer,
+        :machine_type           => node.machineType,
+        :model                  => node.model,
+        :serial_number          => node.serialNumber,
+        :field_replaceable_unit => node.FRU,
+        :host                   => get_host_relationship(node),
+        :power_state            => POWER_STATE_MAP[node.powerStatus],
+        :health_state           => HEALTH_STATE_MAP[node.cmmHealthState.nil? ? node.cmmHealthState : node.cmmHealthState.downcase],
+        :vendor                 => "lenovo",
+        :computer_system        => {
+          :hardware => {
+            :guest_devices => [],
+            :firmwares     => [] # Filled in later conditionally on what's available
+          }
+        },
+        :asset_details          => parse_asset_details(node),
+        :location_led_state    	=> find_loc_led_state(node.leds)
+      }
+      new_result[:computer_system][:hardware] = get_hardwares(node)
+      return node.uuid, new_result
+    end
+
+    def parse_config_pattern(config_pattern)
+      new_result =
+        {
+          :manager_ref  => config_pattern.id,
+          :name         => config_pattern.name,
+          :description  => config_pattern.description,
+          :user_defined => config_pattern.userDefined,
+          :in_use       => config_pattern.inUse
+        }
+      return config_pattern.id, new_result
+    end
+
+    def get_host_relationship(node)
+      # Assign a physicalserver and host if server already exists and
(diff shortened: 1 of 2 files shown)


rodneyhbrown7:
Investigating unintended consequences in the UI after applying this PR.   The physical server with hosts link on the provider summary page throws an exception.

walteraa:
@rodneyhbrown7 could you add the exception here?

rodneyhbrown7:
@miq-bot remove_label wip

AndreyMenezes:
@rodneyhbrown7 I want to understand why you are using service_tag. The API doesn't have any information about it and I can't test with real data.

rodneyhbrown7:
@miq-bot add_label gaprindashvili/yes

walteraa:
@AndreyMenezes `service_tag` is a ManageIQ column which is used as a relationship key between a Host and a Physical server. If you are referring to XClarity API, you should test the `serialNumber` property.

AndreyMenezes:
Thank you @walteraa, I know about the serialNumber from LXCA. I didn't find any information about service_tag, I just want to know what this variable is. But I understand now.

walteraa:
Sure @AndreyMenezes! :smile:

app/models/manageiq/providers/lenovo/physical_infra_manager/refresh_parser.rb
```diff
@@ -318,7 +318,8 @@ def parse_config_pattern(config_pattern)
     def get_host_relationship(node)
       # Assign a physicalserver and host if server already exists and
       # some host match with physical Server's serial number
-      Host.find_by(:service_tag => node.serialNumber)
+      Host.find_by(:service_tag => node.serialNumber) ||
+        Host.joins(:hardware).find_by('hardwares.serial_number' => node.serialNumber)
```
agrare: It looks like there is an inconsistency with the different parsers, vmware sets service_tag but ovirt and scvmm do not.

agrare: @rodneyhbrown7 I updated the SCVMM provider to put the `PhysicalMachine/SMBiosGUID` in `hardware.serial_number` and @jameswnl is in the process of adding the VMware Host's UUID to the same place.

agrare: @rodneyhbrown7 okay VMware has been updated https://github.com/ManageIQ/manageiq-providers-vmware/pull/139 so every infra provider should put the bios uuid in `hardware.serial_number`

Can you do a `where` instead of a find_by?  It is possible that there are duplicates since this comes from host firmware

chargio:
@AndreyMenezes Can you address the comments?

Thanks,
Sergio

app/models/manageiq/providers/lenovo/physical_infra_manager/refresh_parser.rb
AndreyMenezes: @rodneyhbrown7 could you change `hardwares.serial_number` to `hardware.serial_number` like @agrare said? And use `where` instead of `find_by`.

rodneyhbrown7: I am testing this and having problems successfully making the association.   I will continue to evaluate the association here.

chargio:
@AndreyMenezes @rodneyhbrown7 Are we still waiting for updates on this?

app/models/manageiq/providers/lenovo/physical_infra_manager/refresh_parser.rb
```diff
@@ -335,7 +335,8 @@ def parse_config_pattern(config_pattern)
     def get_host_relationship(node)
       # Assign a physicalserver and host if server already exists and
       # some host match with physical Server's serial number
-      Host.find_by(:service_tag => node.serialNumber)
+      Host.find_by(:service_tag => node.serialNumber) ||
```
agrare: We shouldn't need to check both the service_tag and the serial_number now, just the serial_number should be enough.

app/models/manageiq/providers/lenovo/physical_infra_manager/refresh_parser.rb
```diff
@@ -335,7 +335,8 @@ def parse_config_pattern(config_pattern)
...
+        Host.joins(:hardware).where('hardwares.serial_number' => node.serialNumber).take
```
agrare: Up to you how you want to handle it but my thought with using `.where` was we can check if `results.count > 1` and return nil so we don't have a false positive link.

app/models/manageiq/providers/lenovo/physical_infra_manager/refresh_parser.rb
rodneyhbrown7: serial number does not seem to have the same value in all cases so it is not guaranteed to always match.    VMware and RHEVM do not seem to have the same value for this field. I would like to leave this check in.

agrare: What values do they have?  Should be the BIOS UUID in both cases and if not we should fix it.

app/models/manageiq/providers/lenovo/physical_infra_manager/refresh_parser.rb
rodneyhbrown7: using find_by would automatically return a nil if the member was not found.   I will change this back to a find_by selector.  This will make rubocop happy as well.

agrare: I was more worried about if there are duplicate BIOS UUIDs find_by would happily return the first one and we would have no idea that we might be linking to the wrong host.

app/models/manageiq/providers/lenovo/physical_infra_manager/refresh_parser.rb
rodneyhbrown7: If I look at  RHEVM this field this is still a serial number field and we are able to match that to the serial number of the server.      That is what we are matching to find the RHEVM systems.      I think we do want this field to always match the serial number of the server it is sitting on.  The BIOS ID should be captured as a different field in the hardware object.

agrare: So RHEV has the serial number in `hardware.serial_number`, what do you need the `service_tag` for?

app/models/manageiq/providers/lenovo/physical_infra_manager/refresh_parser.rb
agrare: I still think we have an issue with linking to the wrong host if there are duplicate serial numbers

rodneyhbrown7: I am not opposed to resolving that issue, I just think that it would be unexpected to see the BIOS id in the serial number field.

rodneyhbrown7: Removing WIP.   We will work on a second PR to that will look for uuid or serial number in the hardwares.serial_number column.  This PR will enable RHEVM hosts to be discovered.

@miq-bot remove_label wip

rodneyhbrown7:
@miq-bot remove_label wip

simaishi:
@rodneyhbrown7 @agrare https://github.com/ManageIQ/manageiq-providers-lenovo/pull/98 (`gaprindashvili/no`) refactored this part of the code.  As is, the only change that's going to gaprindashvili branch by cherry-picking this MR is:

```diff
--- a/app/models/manageiq/providers/lenovo/physical_infra_manager/refresh_parser.rb
+++ b/app/models/manageiq/providers/lenovo/physical_infra_manager/refresh_parser.rb
@@ -327,7 +327,8 @@ module ManageIQ::Providers::Lenovo
     def get_host_relationship(node)
       # Assign a physicalserver and host if server already exists and
       # some host match with physical Server's serial number
-      Host.find_by(:service_tag => node.serialNumber)
+      Host.find_by(:service_tag => node.serialNumber) ||
+        Host.joins(:hardware).find_by('hardwares.serial_number' => node.serialNumber)
     end
```

Does this look right?

walteraa:
@simaishi I am not sure if it should be `hardwares.serial_number` or `hardware.serial_number`. Ignoring it, I believe it could get the Host through hardware entity when `:service_tag` is empty.

What do you think, @rodneyhbrown7? It looks ok to me.

simaishi:
Marking as `conflict` until I get confirmation...

chargio:
Adding @loicavenel to follow up
@rodneyhbrown7 Have you looked at this?

Thanks,
Sergio

rodneyhbrown7:
@chargio @simaishi  The two line change merged into gaprindashvili is good to go.

simaishi:
Gaprindashvili backport details:

```
$ git log -1
commit fd36d33597b6a359bfd34331567df297857e970b
Author: Adam Grare <agrare@redhat.com>
Date:   Fri Jan 19 10:04:40 2018 -0500

    Merge pull request #93 from rodneyhbrown7/rhevm
    
    searches the hardware resource for a relationship
    (cherry picked from commit 3184e10967025e766665d7192aea1a8364e34847)
    
    Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1552323
```

PR #94, opened 2017-10-26
Add port to new_ems in discover method

lucashsilva:
This PR adds port to new_ems created in discover method. This was necessary because the port was not being included in the provider model, causing an error in the UI when trying to edit any discovered provider.

### DIFF
--- a/app/models/manageiq/providers/lenovo/manager_mixin.rb
+++ b/app/models/manageiq/providers/lenovo/manager_mixin.rb
@@ -88,7 +88,8 @@ module ManageIQ::Providers::Lenovo::ManagerMixin
         new_ems = create!(
           :name     => "Discovered Provider ##{count + 1}",
           :hostname => URI('https://' + ip_address),
-          :zone     => Zone.default_zone
+          :zone     => Zone.default_zone,
+          :port     => port
         )
 
         # Set empty authentications


app/models/manageiq/providers/lenovo/manager_mixin.rb
```diff
@@ -88,7 +88,8 @@ def discover(ip_address, port)
         new_ems = create!(
           :name     => "Discovered Provider ##{count + 1}",
           :hostname => URI('https://' + ip_address),
```
walteraa: I know it isn't a change made by your PR, but could you change this line to use `URI::HTTPS.build`?

rodneyhbrown7: Actually we don't require the protocol prefix at all any more, I think we should remove the 'https://' altogether and just pass in the ip_address into this method.

rodneyhbrown7: The above change can be a separate PR though.   This LGTM.

rodneyhbrown7:
Adding bugzilla ticket open for this issue:  https://bugzilla.redhat.com/show_bug.cgi?id=1505610

@miq-bot add_label gaprindashvili/yes

app/models/manageiq/providers/lenovo/manager_mixin.rb
walteraa: Sounds good for me

simaishi:
This PR was merged before the gaprindashvili branch was created.

PR #95, opened 2017-10-26
Add support for applying a config pattern

skovic:
Added the apply_config_pattern method to the physical server operations module and created a physical server provision task.

### DIFF
--- a/app/models/manageiq/providers/lenovo/physical_infra_manager/operations.rb
+++ b/app/models/manageiq/providers/lenovo/physical_infra_manager/operations.rb
@@ -41,6 +41,10 @@ module ManageIQ::Providers::Lenovo::PhysicalInfraManager::Operations
     change_resource_state(:power_restart_node_controller, args, options)
   end
 
+  def apply_config_pattern(args, options = {})
+    change_resource_state(:deploy_config_pattern, args, options)
+  end
+
   private
 
   def change_resource_state(verb, args, options = {})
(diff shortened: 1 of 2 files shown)


skovic:
@agrare @rodneyhbrown7 Could you guys take a look at this PR? Thanks

app/models/manageiq/providers/lenovo/physical_infra_manager/operations.rb
```diff
@@ -54,9 +58,16 @@ def change_resource_state(verb, args, options = {})
...
 
-    # Turn on the location LED using the xclarity_client API
-    client.send(verb, args.ems_ref)
+    # Execute the action via the client
+    response = case verb
+               when :deploy_config_pattern
+                 client.send(verb, options[:id], [options[:uuid]], options[:restart], options[:etype])
+               else
+                 client.send(verb, args.ems_ref)
+               end
```
agrare: Hm this feels wrong, what if we add another verb which doesn't match these arguments?
Maybe we should just yield the connection and let the caller define the arguments?

skovic: @agrare Regarding how long the action takes, here we are just calling the client which sends a REST call to an LXCA instance to perform the action. The LXCA instance creates a task (in LXCA) and returns a REST response indicating if the request was successfully processed or not. Just as in the case of power operations (which can take awhile), we are not polling the task.

agrare: Perfect @skovic that's exactly what I was wondering, since this is in a provision request we probably do want to poll on the task.

rodneyhbrown7: So I think the undesireable thing here is the insertion of the case statement that does not have a good default. If someone calls another verb, we don't know if it should take the else clause.   As an alternative I recommend that we refactor the method appy_config_pattern method to make its own send call.  We can create a client_connection method that will create the connection and then  each of the public methods can define what method it wants to call on that connection directly.  I think change_resource_state does not add value now that we have broken the pattern of how the client is invoked.

skovic: @agrare I updated the code based on @rodneyhbrown7 's suggestion. Will this work?

Fryguy:
Merging... @agrare if you have any further objections, let's do it in a follow up PR.

PR #98, opened 2017-10-30
Refactoring RefreshParser structure

douglasgabriel:
This PR is able to:
- Refactore the file refresh_parser.rb to a structure that make easier to support other versions of LXCA;
  - Adds a Parser class to provide parse methods with common data at all versions of LXCA, and for all diverging data need to implement a subclass that know how to parse this data;
  - Adds a Dictionary class to provide maps to translate some data.

Depends on:
https://github.com/lenovo/xclarity_client/pull/66


spec/models/manageiq/providers/lenovo/physical_infra_manager/event_catcher/runner_spec.rb
```diff
@@ -13,6 +13,6 @@
   end
 
   it 'will stop the event monitor without any exceptions occurring' do
-    runner.stop_event_monitor
+    runner.stop_event_monitor
```
walteraa: Remove blank spaces.

app/models/manageiq/providers/lenovo/physical_infra_manager/parsers/parser.rb
```diff
@@ -0,0 +1,268 @@
+# Class that provides methods to parse LXCA data to ManageIQ data
+# This class provides only data present in all supported versions
+# Data specific of some version must be parsed on subclasses
+module ManageIQ::Providers::Lenovo
+  class Parser
+    require_relative 'parser_dictionary_constants'
```
walteraa: @douglasgabriel Once it has different client version, doesn't make this line being required inside the versioned parser? If some information meaning (e.g.: `POWER_STATE_MAP` codes) change in the LXCA REST API, in a future version, it can break the parser.

douglasgabriel: Agree. So I've created a method on specific parser that provides the specific dictionary. See [parser_1_3.rb](https://github.com/ManageIQ/manageiq-providers-lenovo/pull/98/files#diff-3c836210612c4d26299119496e46dd2d)

walteraa: Sounds good! :D

AndreyMenezes:
Not merge this PR for now. Refactoring is not safe enough to enter this release.

walteraa:
I totally agree with you. Then don't add the label `gaprindashvili/yes`.

walteraa:
@miq-bot add_label wip

walteraa:
@douglasgabriel I'm not sure but I believe Travis fails because the environment is using RubyGems, the last Rubygems version is `0.5.7 - October 26, 2017 (114 KB)`, then this will work only after a new release in RubyGems. 

@rodneyhbrown7 Do you know when another version will be created? Maybe After this PR is merged https://github.com/lenovo/xclarity_client/pull/62 you could create a `0.5.8` version in RubyGems.

walteraa:
@miq-bot add_label wip

walteraa:
@douglasgabriel It needs rebase

app/models/manageiq/providers/lenovo/physical_infra_manager/parsers/parser_dictionary_constants_2_0.rb
```diff
@@ -0,0 +1,9 @@
+module ManageIQ::Providers::Lenovo
+  # dictionary homologated for versions:
+  # 2.0
+  class ParserDictionaryConstants2X0 < ManageIQ::Providers::Lenovo::ParserDictionaryConstants
+    PROPERTIES_MAP = {
+      :led_identify_name => ["Identify", "Indetification"]
```
walteraa: You should use %(Identify Identification) instead of ["Identify", "Indetification"]. This statement will return a String Array.

It is really "Indetification"? Shouldn't be "Identification"?

douglasgabriel: I Got it. My bad, it is really "Identification", thanks.

walteraa: No problem @douglasgabriel, just a typo! :smile:

walteraa:
@rodneyhbrown7 could you update the xclarity client in RubyGems? After that @douglasgabriel should update the `manageiq-providers-lenovo.gemspec` file and update the xclarity version at the following line `s.add_dependency "xclarity_client", "~> 0.5.3"`. It probably will fix the Travis CI errors.

walteraa:
@douglasgabriel you should remove the [wip] tag only when travis is ok. Could you ask for someone update the xclarity ruby client in rubygems?

walteraa:
@douglasgabriel could you check this offense deteckted by Linter? After that it looks like good to merge.

PR #99, opened 2017-10-31
[GAPRINDASHVILI] We support 2.3.1 on gaprindashvili appliances

jrafanie:
2.4.x was tested on master and is no longer tested on the gaprindashvili


jrafanie:
@miq-bot add_label developer
@miq-bot add_label test

PR #107, opened 2017-11-06
Change the way device type is determined

skovic:
Instead of using the product name to determine the device type, this PR updates the _get_device_type_ method to determine the device type by looking at the port type.


skovic:
@miq-bot add_label gaprindashvili/no

skovic:
@miq-bot assign @rodneyhbrown7

PR #108, opened 2017-11-07
Fix exception handing for credential validation on raw_connect

tumido:
When provider is added the credentials validation is newly done via `raw_connect` (class method) instead of `verify_credentials` (needed EMS object). Further reading on this change can be found here: ManageIQ/manageiq-ui-classic#1580

This change ensures that exceptions are handled properly, because the `verify_credentials` is not called anymore.

Related: https://github.com/ManageIQ/manageiq-ui-classic/pull/2650
Similar PRs for other providers:
- https://github.com/ManageIQ/manageiq-providers-azure/pull/161
- https://github.com/ManageIQ/manageiq-providers-hawkular/pull/100
- https://github.com/ManageIQ/manageiq-providers-scvmm/pull/41
- https://github.com/ManageIQ/manageiq-providers-google/pull/28
- https://github.com/ManageIQ/manageiq-providers-nuage/pull/40

### DIFF
--- a/app/models/manageiq/providers/lenovo/manager_mixin.rb
+++ b/app/models/manageiq/providers/lenovo/manager_mixin.rb
@@ -50,7 +50,7 @@ module ManageIQ::Providers::Lenovo::ManagerMixin
   end
 
   module ClassMethods
-    def raw_connect(username, password, host, port, auth_type, verify_ssl)
+    def raw_connect(username, password, host, port, auth_type, verify_ssl, validate = false)
       xclarity = XClarityClient::Configuration.new(
         :username   => username,
         :password   => password,
@@ -59,7 +59,11 @@ module ManageIQ::Providers::Lenovo::ManagerMixin
         :auth_type  => auth_type,
         :verify_ssl => verify_ssl
       )
-      XClarityClient::Client.new(xclarity)
+      connection = XClarityClient::Client.new(xclarity)
+
+      connection_rescue_block { validate_connection(connection) } if validate
+
+      connection
     end
 
     def validate_connection(connection)


tumido:
cc @jntullo

rodneyhbrown7:
This fix still allowed me to validate an invalid credential.  Needs work?

tumido:
yes, because lenovo provider has not been switched to the queued credential validation. So if there's been an issue before, it's not addressed. This suites as a preparation (makes the behavior consistent for both validation methods) for the workflow switch introduced here https://github.com/ManageIQ/manageiq-ui-classic/pull/1580. It was later revoked for physical_infra here https://github.com/ManageIQ/manageiq-ui-classic/pull/2643.

tumido:
@rodneyhbrown7 using this PR in current master would not bring any change to the workflow. In order to test it, you need to apply this PR and the dependent one + you need a similar change as we do here for middleware (to enable the new workflow) https://github.com/ManageIQ/manageiq-ui-classic/pull/2660

rodneyhbrown7:
Thanks @tumido, I didn't pick up on that.  I will validate with all of the relevant PRs.

rodneyhbrown7:
Note that without this change: ManageIQ/manageiq-ui-classic#2650, the validation will always pass.

@miq-bot add_label gaprindashvili/yes, bug

simaishi:
Gaprindashvili backport details:

```
$ git log -1
commit e7eb2407aa40b3d6ded685f3a1f7fa946c44489d
Author: Rodney Brown <rodneyhbrown7@users.noreply.github.com>
Date:   Wed Nov 8 01:16:07 2017 -0500

    Merge pull request #108 from tumido/fix_cred_validation
    
    Fix exception handing for credential validation on raw_connect
    (cherry picked from commit 3b265e84073a79da4a3e51d58faa24b684586545)
```

PR #122, opened 2018-01-10
Parsing disk capacity from a physical server

felipedf:
This PR is able to:

 * parse the total disk capacity from storage accessed by a physical server


rodneyhbrown7:
@felipedf Can you fix the few lint issues in this PR? I think this can be merged.

PR #125, opened 2018-01-17
AssetDetails should be AssetDetail

bdunne:
Depends on https://github.com/ManageIQ/manageiq/pull/16833


bdunne:
@miq-bot add_labels gaprindashvili/yes

simaishi:
https://github.com/ManageIQ/manageiq/pull/16833 is `gaprindashvili/no`, so changing this PR to no as well.

PR #126, opened 2018-01-18
Migrate model display names from locale/en.yml to plugin

mzazrivec:
This is a continuation of the work started in https://github.com/ManageIQ/manageiq/pull/16596 where we want to migrate display model names from locale/en.yml into particular models, where we'll use standard gettext. Main goal here is to make this one aspect of our application pluggable.

Core PR: https://github.com/ManageIQ/manageiq/pull/16836


app/models/manageiq/providers/lenovo/physical_infra_manager/physical_server.rb
```diff
@@ -1,4 +1,8 @@
 module ManageIQ::Providers
   class Lenovo::PhysicalInfraManager::PhysicalServer < ::PhysicalServer
   end
+
+  def self.display_name(number = 1)
```
felipedf: This should be defined inside the PhysicalServer Class rather then inside the module

agrare:
@mzazrivec can you rebase?

agrare:
@felipedf did @mzazrivec address your concerns?

mzazrivec:
@agrare Done

mzazrivec:
Can this be merged please? (seeing it's approved)

PR #129, opened 2018-01-25
Removing out of date code

douglasgabriel:
**This PR removes:**
* Out of date code added in this PR: https://github.com/ManageIQ/manageiq-providers-lenovo/pull/93

This code was moved to [parse.rb](https://github.com/ManageIQ/manageiq-providers-lenovo/blob/master/app/models/manageiq/providers/lenovo/physical_infra_manager/parsers/parser.rb) in this PR: https://github.com/ManageIQ/manageiq-providers-lenovo/pull/98


douglasgabriel:
Hey @agrare, could you take a look on this?

douglasgabriel:
@miq-bot add_labels gaprindashvili/yes

agrare:
@douglasgabriel does this need to be in gaprindashvili ?  Is it causing an error in refresh?

douglasgabriel:
Not really @agrare. It just clean this amount of duplicated code

agrare:
👍 thanks were only taking blockers only for the first g release so this will get picked up in g-2

simaishi:
@douglasgabriel  @agrare https://github.com/ManageIQ/manageiq-providers-lenovo/pull/98 isn't marked as `gaprindasvhili/yes`, so I assume we don't need this PR?

agrare:
@simaishi I don't think we need this or https://github.com/ManageIQ/manageiq-providers-lenovo/pull/98 in gaprindashvili since that was refactoring

simaishi:
Thanks Adam, removing the label.

PR #132, opened 2018-02-05
Adding parsing steps for switches

saulotoledo:
This PR adds parsing steps for switches. We are now able to get the lists of switches from LXCA and parse them to store in the ManageIQ database.

**Depends on:**

- [https://github.com/ManageIQ/manageiq-schema/pull/160](https://github.com/ManageIQ/manageiq-schema/pull/160) [Merged]
- [https://github.com/ManageIQ/manageiq-schema/pull/165](https://github.com/ManageIQ/manageiq-schema/pull/165) [Merged]
- [https://github.com/ManageIQ/manageiq-schema/pull/182](https://github.com/ManageIQ/manageiq-schema/pull/182) [Merged]
- [https://github.com/ManageIQ/manageiq/pull/16948](https://github.com/ManageIQ/manageiq/pull/16948) [Merged]


saulotoledo:
@agrare, could you take a look at this PR?

saulotoledo:
@miq-bot assign @agrare

app/models/manageiq/providers/lenovo/physical_infra_manager/refresh_parser.rb
```diff
@@ -56,6 +57,14 @@ def get_physical_servers
       process_collection(nodes, :physical_servers) { |node| @parser.parse_physical_server(node) }
     end
 
+    def get_switches
+      return @all_switches if @all_switches
```
felipedf: Is this line needed?

douglasgabriel: Maybe what you want here is to prevent a second call to xclarity API. As @felipedf said @all_switches is never initialized so this line is pointless, will make more sense if you do something like this:

```ruby
def get_switches
    @all_switches = @connection.discover_switches unless @all_switches

    process_collection(@all_switches, :switches) { |switch| @parser.parse_switch(switch) }
end
```

app/models/manageiq/providers/lenovo/physical_infra_manager/parsers/parser.rb
```diff
@@ -35,7 +35,17 @@ def parse_physical_server(node)
...
+
+      return node.uuid, result
+    end
+
+    def parse_switch(node, ems_id = nil)
+      result = parse(node, dictionary::SWITCH)
+      result[:hardware] = get_switch_hardwares(node)
+
+      # Keep track of the ems where this switch is in.
+      result[:ems_id] = ems_id unless ems_id.nil?
```
douglasgabriel: The `.nil?` is not needed here, you can just do `result[:ems_id] = ems_id if ems_id`

app/models/manageiq/providers/lenovo/physical_infra_manager/refresh_parser.rb
douglasgabriel: Seems like what you want here is to prevent a second call to xclarity API, but as @felipedf said, the `@all_switches` is never initialized, so this line is pointless. Will make more sense if you do something like:

```ruby
def get_switches
    @all_switches = @connection.discover_switches unless @all_switches

    process_collection(@all_switches, :switches) { |switch| @parser.parse_switch(switch) }
end
```

douglasgabriel: Seems like what you want here is to prevent a second call to xclarity client, but as @felipedf said the `@all_switches` is never initialized so this line is pointless. Will make more sense if you do something like this:
```ruby
def get_switches
     @all_switches = @connection.discover_switches unless @all_switches

    process_collection(@all_switches, :switches) { |switch| @parser.parse_switch(switch) }
end
```

saulotoledo: You are right. I am working on that.

app/models/manageiq/providers/lenovo/physical_infra_manager/parsers/parser_dictionary_constants.rb
```diff
@@ -37,6 +37,29 @@ class ParserDictionaryConstants
...
+      :power_state   => 'powerState',
+      :product_name  => 'productName',
+      :part_number   => 'partNumber',
+      :serial_number => 'serialNumber',
+      :description   => 'description',
+      :manufacturer  => 'manufacturer',
+    }.freeze
+
+    SWITCH_PORT = {
+      :device_name => 'portName',
```
skovic: Maybe _port_ should be used here instead of _port_name_. I notice that when I run this code against LXCA 2.0.0, the _port_name_ field is empty, resulting in the entry in the _guest_device_ table having a blank _device_name_. However, the _port_ field returned by LXCA contains the switch interface name such as "Ethernet1/2".

app/models/manageiq/providers/lenovo/physical_infra_manager/parsers/parser_dictionary_constants.rb
```diff
@@ -37,6 +37,29 @@ class ParserDictionaryConstants
...
+      :address     => 'peerMacAddress',
```
skovic: The _peerMacAddress_ field being returned by LXCA 2.0.0 seems to be empty for all ports. As a result, the switch ports in my _guest_device_ table have an empty _address_ field. Is the _peerMacAddress_ field normally populated?

saulotoledo: That is a good question. The last LXCA 2.x instances I had access do not have that data anymore. I think this is particularly important because it is the way we have for acquiring the mac address. I will further investigate.

app/models/manageiq/providers/lenovo/physical_infra_manager/parsers/parser_dictionary_constants.rb
saulotoledo: Thank you for your email about that. As stated there, the [LXCA documentation](http://sysmgt.lenovofiles.com/help/index.jsp?topic=%2Fcom.lenovo.lxca_restapis.doc%2Frest_api_switches_uuid_ports_get.html&resultof="peerMacAddress"%20"peermacaddress"%20) presents the _port_ as the port index, and the name as _portName_. Also, I think we should maintain backward compatibility with previous LXCA versions, so we should still use the _portName_ . This is why I did chose that here. I prefer wait for that discussion before updating the code. What do you think about?

app/models/manageiq/providers/lenovo/physical_infra_manager/parsers/parser_dictionary_constants.rb
saulotoledo: @skovic, I started looking for that, but I am convinced that this is the field we want here, despite that points to the address of the entity attached to the port. There is no data in the LXCA API to fill the address in the ManageIQ. You and @rodneyhbrown7 have more experience with the LXCA platform, so if you think we should change that to another field, let me know. Anyway, I think that if we do not have the proper information here, there is a problem in how the LXCA acquires the data.

app/models/manageiq/providers/lenovo/physical_infra_manager/parsers/parser_dictionary_constants.rb
skovic: Based on Ionut's latest response in the email thread, I believe that we want to use _port_ since he said it always contains the interface name.  Adding @rodneyhbrown7 to the conversation.

app/models/manageiq/providers/lenovo/physical_infra_manager/parsers/parser_dictionary_constants.rb
skovic: Since the model we are creating in the database represents the switch and its ports, I don't think it is correct to use the _peerMacAddress_ as the address of the switch's port because it is the MAC address of the device that's attached to the switch -- not the MAC address of the switch port itself. @rodneyhbrown7 Should we not fill in this field, or can we somehow get the switch port MAC addresses from LXCA?

app/models/manageiq/providers/lenovo/physical_infra_manager/parsers/parser_dictionary_constants.rb
douglasgabriel: Maybe it will be a good idea try to get the name of the port from `portName` and if it is empty, get from `port` once the official documentation says that `portName` really is the field for port name and `port` is "only" the index of the port. This way we provide a compatibility with LXCAs that provides such date in both fields.

saulotoledo: As we discussed in our meeting I will implement as suggested by @douglasgabriel. So we can have both. Thank you, @skovic, for finding that problem.

app/models/manageiq/providers/lenovo/physical_infra_manager/parsers/parser_dictionary_constants.rb
saulotoledo: @skovic, @rodneyhbrown7, as you said in our meeting, I will create a new `peerMacAddress` field in the schema to store that data and I will leave the address blank.

saulotoledo:
@miq-bot remove label unmergeable

saulotoledo:
@miq-bot remove-label unmergeable

saulotoledo:

saulotoledo:
@rodneyhbrown7, @agrare, this is ok for review.

saulotoledo:
@miq-bot remove_tag wip

saulotoledo:
@miq-bot remove_label wip

saulotoledo:

app/models/manageiq/providers/lenovo/physical_infra_manager/parsers/components/firmware_parser.rb
```diff
@@ -0,0 +1,25 @@
+require_relative 'component_parser'
```
Fryguy: Don't use require or require_relative as we should allow Rails to use its built-in autoload functionality.  The main reason is that in dev mode, Rails will reload classes/constants from disk on source change, and using require/require_relative prevents that, which can actually break things.   Additionally, the class in this file does not match the file path, which breaks Rails autoload.

agrare:
@saulotoledo can you address the rubocop issues?  They look like valid issues

app/models/manageiq/providers/lenovo/physical_infra_manager/parsers/components/firmware_parser.rb
saulotoledo: Done by rebasing with @agrare 's PR.

saulotoledo:
@agrare Done. The tests are failing because they are trying to use the `part_number` column added by https://github.com/ManageIQ/manageiq-schema/pull/182.

agrare:
@saulotoledo I get this failure with your schema changes applied:

```
Failures:

  1) ManageIQ::Providers::Lenovo::PhysicalInfraManager::Refresher will perform a full refresh
     Failure/Error: expect(GuestDevice.count).to eq(6)
     
       expected: 6
            got: 38
     
       (compared using ==)
     # ./spec/models/manageiq/providers/lenovo/physical_infra_manager/refresher_spec.rb:104:in `assert_table_counts'
     # ./spec/models/manageiq/providers/lenovo/physical_infra_manager/refresher_spec.rb:76:in `block (3 levels) in <top (required)>'
     # ./spec/models/manageiq/providers/lenovo/physical_infra_manager/refresher_spec.rb:72:in `times'
     # ./spec/models/manageiq/providers/lenovo/physical_infra_manager/refresher_spec.rb:72:in `block (2 levels) in <top (required)>'

Finished in 5.97 seconds (files took 5.24 seconds to load)
70 examples, 1 failure
```

With that expect changed the tests are passing locally

saulotoledo:
@agrare Thanks for that info. It is caused by the full refresh test (maybe we should change that in the future). I've updated the count to match the database. It should work now.

PR #140, opened 2018-03-15
Fixing bug in the calculate of physical server total memory

douglasgabriel:
Bug Fix
======
Once the LXCA treat the memory amount in GB and the MiQ treat as MB, it needs to be converted on parse.

**Actual result**
On LXCA:
error_in_lxca
On MiQ:
error_in_miq

**After this PR fix:**
On MiQ:
error_fix_miq

### DIFF
--- a/app/models/manageiq/providers/lenovo/physical_infra_manager/parsers/parser.rb
+++ b/app/models/manageiq/providers/lenovo/physical_infra_manager/parsers/parser.rb
@@ -125,7 +125,8 @@ module ManageIQ::Providers::Lenovo
     end
 
     def get_memory_info(node)
-      node.memoryModules&.reduce(0) { |total, mem| total + mem['capacity'] }
+      total_memory_gigabytes = node.memoryModules&.reduce(0) { |total, mem| total + mem['capacity'] }
+      total_memory_gigabytes * 1024 # convert to megabytes
     end
 
     def get_total_cores(node)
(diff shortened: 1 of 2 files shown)


douglasgabriel:
@miq-bot add_label bug, gaprindashvili/yes

douglasgabriel:
@miq-bot assign @agrare

simaishi:
@douglasgabriel @agrare Please add BZ link.

douglasgabriel:
@simaishi https://bugzilla.redhat.com/show_bug.cgi?id=1557363

simaishi:
@douglasgabriel `app/models/manageiq/providers/lenovo/physical_infra_manager/parsers/parser.rb` doesn't exist in Gaprindashvili branch as https://github.com/ManageIQ/manageiq-providers-lenovo/pull/98 isn't backported, and `def get_memory_info` in `app/models/manageiq/providers/lenovo/physical_infra_manager/refresh_parser.rb ` is different between `master` and `gaprindashvili`... Can you please create a PR for Gaprindashvili branch?

douglasgabriel:
Actually. @simaishi, this error doesn't exist on gaprindashvili branch, once the operation added by this PR is already being make. [see](https://github.com/ManageIQ/manageiq-providers-lenovo/blob/gaprindashvili/app/models/manageiq/providers/lenovo/physical_infra_manager/refresh_parser.rb#L83)

simaishi:
Ok, will close the gaprindashvili BZ then. Thanks!

PR #145, opened 2018-03-27
Parsing PCI Devices and Addin Cards as Guest Devices

douglasgabriel:
__This PR is able to:__
- Add `NetworkDeviceParser` component;
- Parse Network Devices on PCI devices and Addin Cards.

### DIFF
--- a/app/models/manageiq/providers/lenovo/physical_infra_manager/parsers/components/physical_server_parser.rb
+++ b/app/models/manageiq/providers/lenovo/physical_infra_manager/parsers/components/physical_server_parser.rb
@@ -89,6 +89,9 @@ module ManageIQ::Providers::Lenovo
 
         def get_addin_cards(node)
           parsed_addin_cards = []
+
+          cards_to_parse = select_cards_to_parse(node)
+
           # For each of the node's addin cards, parse the addin card and then see
           # if it is already in the list of parsed addin cards. If it is, see if
           # all of its ports are already in the existing parsed addin card entry.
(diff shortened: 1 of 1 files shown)


douglasgabriel:
@rodneyhbrown7 @skovic @agrare could you take a look on this?

app/models/manageiq/providers/lenovo/physical_infra_manager/parsers/components/network_devices_parser.rb
```diff
@@ -0,0 +1,100 @@
...
+        private
+
+        #
+        # Get all devices with the class `Network controller` in `pciDevices` and `addinCards` field
+        # and remove duplicated devices considering the `pciSubID` field
+        #
+        def select_network_devices(component)
+          devices = []
+          devices << component.try(:pciDevices)&.select { |device| device["class"] == "Network controller" }
+          devices << component.try(:addinCards)&.select { |device| device["class"] == "Network controller" }
```
rodneyhbrown7: I think that the REST API for 1.4.0 does not have this property on it.  You should be able to use the spec tests/vcrs to validate that this will still work.

rodneyhbrown7: Also, do you need both the try and the &.  syntax?  My understanding is that .try()  and  ()&  are equivalent behaviours.

rodneyhbrown7:
Douglas, I found that same cards do not have the tag "Network" in the name of the device even though they are network devices. The previous revision of this PR was able to detect these devices by using the class of the adapter as a key. Can that check still be done?

douglasgabriel:
Good point, I'll look a way to include this verification on this code. Also, the `select` method promote a good clean up but in some cases it was failing, so I try this new approach because maybe it would take a longer time to realize what was missing in the other code. But my plan is soon open a new PR with a functional refactor

douglasgabriel:
done @rodneyhbrown7

PR #147, opened 2018-03-29
Parsing physical rack

felipedf:
Parse info from a PhysicalRack

 * To be merged with: ManageIQ/manageiq#16853

### DIFF
--- a/app/models/manageiq/providers/lenovo/physical_infra_manager.rb
+++ b/app/models/manageiq/providers/lenovo/physical_infra_manager.rb
@@ -1,6 +1,4 @@
 class ManageIQ::Providers::Lenovo::PhysicalInfraManager < ManageIQ::Providers::PhysicalInfraManager
-  has_many :physical_servers, :foreign_key => "ems_id", :class_name => "ManageIQ::Providers::Lenovo::PhysicalInfraManager::PhysicalServer", :dependent => :destroy
-
   include ManageIQ::Providers::Lenovo::ManagerMixin
   include_concern 'Operations'
 
--- a/app/models/manageiq/providers/lenovo/physical_infra_manager/physical_rack.rb
+++ b/app/models/manageiq/providers/lenovo/physical_infra_manager/physical_rack.rb
@@ -1,0 +1,4 @@
+module ManageIQ::Providers
+  class Lenovo::PhysicalInfraManager::PhysicalRack < ::PhysicalRack
+  end
+end
--- a/app/models/manageiq/providers/lenovo/physical_infra_manager/parsers/components/physical_rack_parser.rb
+++ b/app/models/manageiq/providers/lenovo/physical_infra_manager/parsers/components/physical_rack_parser.rb
@@ -1,0 +1,23 @@
+require_relative 'component_parser'
+
+module ManageIQ::Providers::Lenovo
+  module Parsers
+    class PhysicalRackParser < ComponentParser
+      class << self
+        #
+        # Parse a rack object to a hash with its data
+        #
+        # @param cab [XClarity::PhysicalRack] a rack object
+        # @param physical_servers [Hash] a already parsed physical_servers that belong to cab
+        #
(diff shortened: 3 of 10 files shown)


felipedf:
The Travis error is due the server[:rack] field that is not delete after parsing, this is done here ManageIQ/manageiq#16853 and will be ok after merge. @agrare

rodneyhbrown7:
Can you rebase this with master?

felipedf:
Done

felipedf:
@miq-bot assign @agrare

agrare:
@felipedf can you add some tests to the [full refresh test ](https://github.com/felipedf/manageiq-providers-lenovo/blob/082a09a73665d0add56001357e93672cdba64bbd/spec/models/manageiq/providers/lenovo/physical_infra_manager/refresher_spec.rb#L68) to check for the physical racks and the association between the physical servers and the physical racks?

Convention here would be to add a `assert_specific_rack` to check for a specific rack's attributes and a `assert_specific_server` where you can check the server has a relationship back to the rack.

app/models/manageiq/providers/lenovo/physical_infra_manager.rb
```diff
@@ -1,6 +1,8 @@
 class ManageIQ::Providers::Lenovo::PhysicalInfraManager < ManageIQ::Providers::PhysicalInfraManager
-  has_many :physical_servers, :foreign_key => "ems_id", :class_name => "ManageIQ::Providers::Lenovo::PhysicalInfraManager::PhysicalServer", :dependent => :destroy
-
+  has_many :physical_racks, :foreign_key => "ems_id", :dependent => :destroy, :inverse_of => :ext_management_system,
+           :class_name => "ManageIQ::Providers::Lenovo::PhysicalInfraManager::PhysicalRack"
+  has_many :physical_servers, :foreign_key => "ems_id", :dependent => :destroy, :inverse_of => :ext_management_system,
+           :class_name => "ManageIQ::Providers::Lenovo::PhysicalInfraManager::PhysicalServer"
```
agrare: Shouldn't the physical_racks and physical_servers associations be on the core physical_infra_manager?  And you shouldn't need the class name as long as the assoc name matches the base class name

app/models/manageiq/providers/lenovo/physical_infra_manager/physical_server.rb
```diff
@@ -2,6 +2,12 @@ module ManageIQ::Providers
   class Lenovo::PhysicalInfraManager::PhysicalServer < ::PhysicalServer
     include_concern 'RemoteConsole'
 
+    belongs_to :ext_management_system, :foreign_key => :ems_id, :inverse_of => :physical_servers,
+               :class_name => "ManageIQ::Providers::Lenovo::PhysicalInfraManager"
+
+    belongs_to :physical_rack, :dependent => :destroy, :inverse_of => :physical_servers,
+               :class_name => "ManageIQ::Providers::Lenovo::PhysicalInfraManager::PhysicalRack"
```
agrare: Again I think all of these associations should be defined on the base models without the class_name, not here

app/models/manageiq/providers/lenovo/physical_infra_manager/physical_rack.rb
```diff
@@ -0,0 +1,9 @@
+module ManageIQ::Providers
+  class Lenovo::PhysicalInfraManager::PhysicalRack < ::PhysicalRack
+    belongs_to :ext_management_system, :foreign_key => :ems_id, :inverse_of => :physical_racks,
+               :class_name => "ManageIQ::Providers::Lenovo::PhysicalInfraManager"
+
+    has_many :physical_servers, :dependent => :destroy, :inverse_of => :physical_rack,
+             :class_name => "ManageIQ::Providers::Lenovo::PhysicalInfraManager::PhysicalServer"
```
agrare: Same this should be in the base model

PR #151, opened 2018-04-09
Update xclarity_client gem

Sanmoo:
'~> 0.5.x' would not pick the latest version '0.6.0', as it equals to
['>= 0.5.x', '< 0.6.0'].
Reference:
http://guides.rubygems.org/patterns/#pessimistic-version-constraint.


Sanmoo:
@rodneyhbrown7 , I have did some basic tests:

* Verified the new version was picked correctly after running bundle install in manageiq.
* Verified the provider refresh worked.

I believe this PR is good to be merged.

PR #155, opened 2018-04-12
[GAPRINDASHVILI] [BZ 1563796] Fixing the network devices and ports parser

douglasgabriel:
Bug Fix
======
When execute the refresh for the first time all Network Devices and Ports are saved successfully, but, from the second time onwards, some Network Devices are deleted and sometimes the ports are duplicated.

__Actual result:__

`guest_devices` table after first refresh:

image

`guest_devices` table after second refresh:

image

This problem occurs because the devices are distinguish by their type and uuid, and once that before this PR the uid_ems field was not filled, the network devices and its ports were distinguish incorrectly causing a wrong update.

__Bugzilla:__
https://bugzilla.redhat.com/show_bug.cgi?id=1563796


JPrause:
@miq-bot add_label blocker

PR #118, opened 2017-12-07
Pinning faker gem to 1.8.3

juliancheal:
faker changed dependency of i18n which conflicts with our dependency from manageiq-consumption:
stympy/faker@9128945#diff-e79a60dc6b85309ae70a6ea8261eaf95R5

PR #165, opened 2018-04-26
Move faker gem to development

saulotoledo:
Faker gem is only used for specs. This PR moves it to development.
It is pinned as `1.8.3` due to https://github.com/ManageIQ/manageiq-providers-lenovo/pull/118.


saulotoledo:
@miq-bot assign @rodneyhbrown7

PR #168, opened 2018-05-04
Fix bug when portInfo was null

felipedf:
The refresh was crashing when a port didn't contain a "portInfo" field

### DIFF
--- a/app/models/manageiq/providers/lenovo/physical_infra_manager/parser/physical_network_ports_parser.rb
+++ b/app/models/manageiq/providers/lenovo/physical_infra_manager/parser/physical_network_ports_parser.rb
@@ -29,7 +29,7 @@ module ManageIQ::Providers::Lenovo
 
       def parse_physical_server_ports(port)
         port_info = port["portInfo"]
-        physical_ports = port_info["physicalPorts"]
+        physical_ports = port_info&.dig('physicalPorts')
         physical_ports&.map do |physical_port|
           parsed_physical_port = parse_physical_port(physical_port)
           logical_ports = physical_port["logicalPorts"]


felipedf:
@miq-bot add_label bug

felipedf:
@miq-bot assign @agrare 
@rodneyhbrown7

app/models/manageiq/providers/lenovo/physical_infra_manager/parser/network_device_parser.rb
```diff
@@ -90,7 +90,7 @@ def parse_child_devices(device, all_devices)
 
       def parse_ports(port)
         port_info = port["portInfo"]
-        physical_ports = port_info["physicalPorts"]
+        physical_ports = port_info&.try(:[], 'physicalPorts')
```
douglasgabriel: I believe that a `try` call here is enough, you can remove the `&`.

app/models/manageiq/providers/lenovo/physical_infra_manager/parser/network_device_parser.rb
```diff
@@ -90,7 +90,7 @@ def parse_child_devices(device, all_devices)
...
         physical_ports&.map do |physical_port|
```
douglasgabriel: To keep the pattern, could you replace the `&` with a `try` call?

app/models/manageiq/providers/lenovo/physical_infra_manager/parser/network_device_parser.rb
agrare: That's correct, or you can use `port_info&.dig('physicalPorts')`...up to you but I've never liked the way `.try` looks with a hash dereference

felipedf: Cool, .dig seems cleaner to me

agrare:
@felipedf looks good but can you rebase on master and resolve the conflict?

PR #169, opened 2018-05-07
Refactoring the management network table to add consistency

felipedf:
Parsing a physical server's network info trough `hardware` instead through `guest_device`.
Add a parse of a `subnet` and `default_gateway`


felipedf:
Yes @skovic it is still WIP we haven't decided whether or not we will actually change this table.
and you are right it should have a refactoring label, so adding it.

PR #172, opened 2018-05-16
Parser refactoring

caiocmpaes:
**This PR makes the following refactoring:**
- Moved ipaddress and hostname updates to physical_infra_manager
- Using local variable to return inventory
- Removing unused refresh_helper_methods
- Passing only connection info to refresh_parser
- Fix Rubocop issues and force single-quote strings
- Removing unused miq_template_type class method
- Simple change in component_parser to allow upcoming changes
- Common Maps were moved to the parent class component_parser
- Moved every map definition to the parser class that uses It
- Switches ports are different from server ports. Different parsers then.
- Extract a management_device_parser
- Removing some parsers and adding their logic inside existing ones
- Computed properties can be declared through Parser Maps
- Extract a superclass GuestDeviceParser from storage and network

All the work was split into each commit. I tried to let the changes more trackable and avoid changes at the tests for the first moment. Only string quote changes were applied on the tests. On a later time I intend to make some refactoring & enhancement in the tests.

# Important
Please, review this code through each commit, clicking on the oldest commit and then `next` to go to the others. I tried to put context changes inside each commit. Also, there are comments on each commit message.


caiocmpaes:
@miq-bot assign @agrare 
/cc @douglasgabriel @felipedf @CharlleDaniel @saulotoledo

caiocmpaes:
I am closing this PR in order to make It easier to review. I am going to split this commits into self-contained Pull Requests.

PR #174, opened 2018-05-17
Removing type assigment from PhysicalSwitchParser

EsdrasVP:
This PR is able to:
- Remove unnecessary type assigment from PhysicalSwitchParser, since the migration already covers that

Depends on:
- [ManageIQ/manageiq-schema#202](https://github.com/ManageIQ/manageiq-schema/pull/202)
- [ManageIQ/manageiq#17431](https://github.com/ManageIQ/manageiq/pull/17431)


EsdrasVP:
This was resolved with a better solution, so I'll close this PR.

PR #177, opened 2018-05-22
Adding physical switch power operations

douglasgabriel:
__This PR is able to:__
- Add the physical switch power operation;
- Add specs for it;

__Depends on:__
~https://github.com/lenovo/xclarity_client/pull/112~ - Merged

### DIFF
--- a/app/models/manageiq/providers/lenovo/physical_infra_manager/physical_switch/operations.rb
+++ b/app/models/manageiq/providers/lenovo/physical_infra_manager/physical_switch/operations.rb
@@ -1,0 +1,37 @@
+#
+# Has the Power Operation methods for Physical Switches
+#
+module ManageIQ::Providers::Lenovo::PhysicalInfraManager::PhysicalSwitch::Operations
+  extend ActiveSupport::Concern
+
+  #
+  # Restarts the physical switch.
+  #   it does the `power_cycle_soft` operation.
+  #
+  def restart
+    change_resource_state(:power_cycle_soft_switch)
+  end
+
+  private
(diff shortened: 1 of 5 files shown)


spec/models/manageiq/providers/lenovo/physical_infra_manager/physical_switch_spec.rb
```diff
@@ -0,0 +1,39 @@
+require "xclarity_client"
+require "faker"
```
saulotoledo: Remove unused imports.

app/models/manageiq/providers/lenovo/physical_infra_manager/physical_switch/operations.rb
```diff
@@ -0,0 +1,34 @@
...
+  private
+
+  #
+  # Sends the power operation for a Switch.
+  #
+  # @param [symbol] verb - the operation that must be sent
+  #
+  # @return the LXCA response
+  #
+  def change_resource_state(verb)
```
agrare: This looks really similar to the method in [PhysicalInfraManager::Operations](https://github.com/ManageIQ/manageiq-providers-lenovo/blob/master/app/models/manageiq/providers/lenovo/physical_infra_manager/operations.rb#L60), I wonder if we can generalize that?  Maybe some future refactoring

douglasgabriel: You're right, but I agree with do this refactoring in a separate PR

PR #178, opened 2018-05-22
Use local variable to return inventory (#2)

caiocmpaes:
**What this PR does:**
- Makes a **refactoring** on the `RefreshParser` logic and removes the use of the instance variable `@data` to return EMS inventory.

**Reason:**
- The `process_collection` method simply populates the `data` instance variable. We can return these results without the need to use variable in class scope.

**Depends on:**
- ~#173~ [merged]

PR #173, opened 2018-05-17
Remove ipaddress and hostname updates from physical_infra_manager (#1)

caiocmpaes:
**What this PR does:**
- Removes the updates of hostname and IP Address features.

**Reason:**

- I believe the `RefreshParser` responsibility is to handle the inventory fetching and parsing. Also, the updates are unnecessary and may create errors in some use cases.

PR #179, opened 2018-05-22
Removing RefreshHelperMethods unused code (#3)

caiocmpaes:
**What this PR does:**
- Removes the unused code inside `RefreshHelperMethods` module.
- Removes unnecessary instance variables inside `RefreshParser#initialize`.

**Depends on:**
- ~#178~ [Merged]

### DIFF
--- a/app/models/manageiq/providers/lenovo/physical_infra_manager/refresh_parser.rb
+++ b/app/models/manageiq/providers/lenovo/physical_infra_manager/refresh_parser.rb
@@ -7,19 +7,15 @@ module ManageIQ::Providers::Lenovo
       parent::Parser::ParserDictionaryConstants::MIQ_TYPES["template"]
     end
 
-    def initialize(ems, options = nil)
+    def initialize(ems, _options = nil)
       ems_auth = ems.authentications.first
 
-      @ems               = ems
-      @connection        = ems.connect(:user => ems_auth.userid,
-                                       :pass => ems_auth.password,
-                                       :host => ems.endpoints.first.hostname,
-                                       :port => ems.endpoints.first.port)
-      @options           = options || {}
-      @data              = {}
-      @data_index        = {}
-      @host_hash_by_name = {}
-      @parser            = init_parser(@connection)
+      @ems        = ems
+      @connection = ems.connect(:user => ems_auth.userid,
+                                :pass => ems_auth.password,
+                                :host => ems.endpoints.first.hostname,
+                                :port => ems.endpoints.first.port)
+      @parser     = init_parser(@connection)
     end
 
     def ems_inv_to_hashes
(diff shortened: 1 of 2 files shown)


caiocmpaes:

rodneyhbrown7:
LGTM  _ waiting for resolution of comments on https://github.com/ManageIQ/manageiq-providers-lenovo/pull/173

caiocmpaes:
@agrare this is the next (third) PR on the refactoring I did.

app/models/manageiq/providers/lenovo/physical_infra_manager/refresh_parser.rb
```diff
@@ -1,25 +1,19 @@
 # rubocop:disable Naming/AccessorMethodName
 module ManageIQ::Providers::Lenovo
   class PhysicalInfraManager::RefreshParser < EmsRefresh::Parsers::Infra
-    include ManageIQ::Providers::Lenovo::RefreshHelperMethods
-
     def self.miq_template_type
       parent::Parser::ParserDictionaryConstants::MIQ_TYPES["template"]
     end
 
-    def initialize(ems, options = nil)
```
agrare: Why did you get rid of options?  Even if it isn't used today it might be helpful in the future.

app/models/manageiq/providers/lenovo/physical_infra_manager/refresher.rb
```diff
@@ -4,7 +4,7 @@ def parse_legacy_inventory(ems)
       log_header = "MIQ_LENOVO(#{self.class.name}.#{__method__} Calling for [#{ems.name}])"
       $log.info("#{log_header}")
 
-      ManageIQ::Providers::Lenovo::PhysicalInfraManager::RefreshParser.ems_inv_to_hashes(ems, refresher_options)
+      ManageIQ::Providers::Lenovo::PhysicalInfraManager::RefreshParser.new(ems).ems_inv_to_hashes
```
agrare: All of the other providers follow this pattern, it might not be strictly necessary here but it is better to keep things consistent.

app/models/manageiq/providers/lenovo/physical_infra_manager/refresh_parser.rb
caiocmpaes: I just removed because It is unused. I will set the variable as `_options` then.

app/models/manageiq/providers/lenovo/physical_infra_manager/refresher.rb
caiocmpaes: @agrare What pattern do you want me to keep? passing two variables, or using a class method instead of instance method?

agrare: Both :smile:

caiocmpaes:
@agrare I did the modifications to let the code consistent with the other providers.

PR #180, opened 2018-05-22
Passing only connection info to RefreshParser (#4)

caiocmpaes:
**What this PR does:**
- Add a `connection` method to the `PhysicalInfraManager` so that It can return how to connect to the external provider;
- Changes the initialization of `RefreshParser` to accept only the connection to an external provider and not all the info that a ems has to offer.
- Updates the tests to pass the connection instead of the entire `ems` object.

**Reason:**
- The `RefreshParser` should only require initialization parameters that are important to its execution. Thus, It only needs a connection object.

**Depends on:**
- ~#179~ [MERGED]

### DIFF
--- a/app/models/manageiq/providers/lenovo/physical_infra_manager.rb
+++ b/app/models/manageiq/providers/lenovo/physical_infra_manager.rb
@@ -19,4 +19,12 @@ class ManageIQ::Providers::Lenovo::PhysicalInfraManager < ManageIQ::Providers::P
   def self.description
     @description ||= "Lenovo XClarity"
   end
+
+  # Returns a new connection to the LXCA
+  def connection
+    connect(:user => authentications.first.userid,
+            :pass => authentications.first.password,
+            :host => endpoints.first.hostname,
+            :port => endpoints.first.port)
+  end
 end
--- a/app/models/manageiq/providers/lenovo/physical_infra_manager/refresh_parser.rb
+++ b/app/models/manageiq/providers/lenovo/physical_infra_manager/refresh_parser.rb
@@ -7,27 +7,15 @@ module ManageIQ::Providers::Lenovo
       parent::Parser::ParserDictionaryConstants::MIQ_TYPES["template"]
     end
 
-    def initialize(ems, _options = nil)
-      ems_auth = ems.authentications.first
-
-      @ems        = ems
-      @connection = ems.connect(:user => ems_auth.userid,
-                                :pass => ems_auth.password,
-                                :host => ems.endpoints.first.hostname,
-                                :port => ems.endpoints.first.port)
-      @parser     = init_parser(@connection)
+    def initialize(connection, _options = nil)
+      @connection = connection
+      @parser     = init_parser
     end
 
     def ems_inv_to_hashes
-      log_header = "MIQ(#{self.class.name}.#{__method__}) Collecting data for EMS : [#{@ems.name}] id: [#{@ems.id} ref: #{@ems.uid_ems}]"
-
-      $log.info("#{log_header}...")
-
       inventory                         = get_all_physical_infra
       inventory[:physical_switches]     = get_physical_switches
       inventory[:customization_scripts] = get_config_patterns
-
-      $log.info("#{log_header}...Complete")
 
       inventory
     end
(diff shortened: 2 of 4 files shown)


caiocmpaes:

caiocmpaes:

caiocmpaes:

EsdrasVP:
I think that if the previous PR is ok, this one should be fine.

caiocmpaes:
@miq-bot assign @agrare

app/models/manageiq/providers/lenovo/physical_infra_manager.rb
```diff
@@ -19,4 +19,12 @@ def self.ems_type
   def self.description
     @description ||= "Lenovo XClarity"
   end
+
+  # Returns a new connection to the LXCA
+  def connection
+    connect(:user => authentications.first.userid,
+            :pass => authentications.first.password,
+            :host => endpoints.first.hostname,
+            :port => endpoints.first.port)
```
agrare: If all this does is allow you to call connect with the default values why not just call connect and let it use the defaults it already has?
It has the same defaults for all of these, https://github.com/ManageIQ/manageiq-providers-lenovo/blob/master/app/models/manageiq/providers/lenovo/manager_mixin.rb#L18-L21

app/models/manageiq/providers/lenovo/physical_infra_manager/refresh_parser.rb
```diff
@@ -7,36 +7,24 @@ def self.miq_template_type
...
-      @parser     = init_parser(@connection)
+    def initialize(connection, _options = nil)
+      @connection = connection
+      @parser     = init_parser
     end
 
     def ems_inv_to_hashes
-      log_header = "MIQ(#{self.class.name}.#{__method__}) Collecting data for EMS : [#{@ems.name}] id: [#{@ems.id} ref: #{@ems.uid_ems}]"
-
-      $log.info("#{log_header}...")
```
agrare: Why remove these logs?  They are actually extremely useful for debugging by support.

app/models/manageiq/providers/lenovo/physical_infra_manager/refresher.rb
```diff
@@ -1,10 +1,15 @@
 module ManageIQ::Providers::Lenovo
   class PhysicalInfraManager::Refresher < ManageIQ::Providers::BaseManager::Refresher
     def parse_legacy_inventory(ems)
-      log_header = "MIQ_LENOVO(#{self.class.name}.#{__method__} Calling for [#{ems.name}])"
-      $log.info("#{log_header}")
+      $log.info("MIQ_LENOVO(#{self.class.name}.#{__method__} Calling for [#{ems.name}])")
 
-      ManageIQ::Providers::Lenovo::PhysicalInfraManager::RefreshParser.ems_inv_to_hashes(ems, refresher_options)
+      # Perform a full fetch of the EMS inventory
+      log_header = "Collecting data for EMS : [#{ems.name}] id: [#{ems.id} ref: #{ems.uid_ems}]"
```
agrare: Oh I see you moved them here, we have them in ems_inv_to_hashes because some providers collect and parse in two different steps.  This helps isolate issues with connectivity to the provider, I would recommend leaving them there so that we could do that in the future.

app/models/manageiq/providers/lenovo/physical_infra_manager/refresh_parser.rb
```diff
@@ -7,36 +7,24 @@ def self.miq_template_type
...
-    def initialize(ems, _options = nil)
-      ems_auth = ems.authentications.first
-
-      @ems        = ems
-      @connection = ems.connect(:user => ems_auth.userid,
-                                :pass => ems_auth.password,
-                                :host => ems.endpoints.first.hostname,
-                                :port => ems.endpoints.first.port)
-      @parser     = init_parser(@connection)
+    def initialize(connection, _options = nil)
```
agrare: :-1: for changing the initializer, this is a common pattern across all refresh_parsers.

caiocmpaes:
@agrare I agree with all your comments.

The only motivation for this small refactoring was to pass only the connection data to the parser, as It is what is used inside the `refresh_parser` now. However, in order to maintain consistency with the other providers and allow future use of the EMS object, I am going to change nothing there.

Closing this PR ...

PR #181, opened 2018-05-22
Fix all rubocop issues (#5)

caiocmpaes:
**What this PR does:**
- Fix all rubocop issues;
- Removes this generated paths from rubocop lint:
  - lib/manageiq-providers-lenovo.rb
  - Rakefile
  - manageiq-providers-lenovo.gemspec
  - bin/**

**Reason:**
- Running `rubocop` on the project root will show new problems with added code.

**Depends on:**
- ~#179~ [MERGED]

### DIFF
--- a/.rubocop_local.yml
+++ b/.rubocop_local.yml
@@ -3,3 +3,11 @@ GlobalVars:
   # Loggers
   - $lenovo_log
   - $log
+
+# Ignores the file at lib folder that needs to have a naming with dashes.
+AllCops:
+  Exclude:
+  - lib/manageiq-providers-lenovo.rb
+  - Rakefile
+  - manageiq-providers-lenovo.gemspec
+  - bin/**
--- a/app/models/manageiq/providers/lenovo/manager_mixin.rb
+++ b/app/models/manageiq/providers/lenovo/manager_mixin.rb
@@ -74,7 +74,7 @@ module ManageIQ::Providers::Lenovo::ManagerMixin
 
     def connection_rescue_block
       yield
-    rescue => err
+    rescue StandardError => err
       miq_exception = translate_exception(err)
       raise unless miq_exception
 
@@ -116,7 +116,7 @@ module ManageIQ::Providers::Lenovo::ManagerMixin
         _log.info("Created EMS: #{new_ems.name} with id: #{new_ems.id}")
       end
 
-      EmsRefresh.queue_refresh(new_ems) unless new_ems.blank?
+      EmsRefresh.queue_refresh(new_ems) if new_ems.present?
     end
 
     def discover_queue(ip_address, port, zone = nil)
--- a/app/models/manageiq/providers/lenovo/manager_mixin.rb
+++ b/app/models/manageiq/providers/lenovo/manager_mixin.rb
@@ -74,7 +74,7 @@ module ManageIQ::Providers::Lenovo::ManagerMixin
 
     def connection_rescue_block
       yield
-    rescue StandardError => err
+    rescue => err
       miq_exception = translate_exception(err)
       raise unless miq_exception
 
--- a/app/models/manageiq/providers/lenovo/physical_infra_manager/refresher.rb
+++ b/app/models/manageiq/providers/lenovo/physical_infra_manager/refresher.rb
@@ -2,7 +2,7 @@ module ManageIQ::Providers::Lenovo
   class PhysicalInfraManager::Refresher < ManageIQ::Providers::BaseManager::Refresher
     def parse_legacy_inventory(ems)
       log_header = "MIQ_LENOVO(#{self.class.name}.#{__method__} Calling for [#{ems.name}])"
-      $log.info("#{log_header}")
+      $log.info(log_header)
 
       ManageIQ::Providers::Lenovo::PhysicalInfraManager::RefreshParser.ems_inv_to_hashes(ems, refresher_options)
     end
--- a/app/models/manageiq/providers/lenovo/physical_infra_manager/refresher.rb
+++ b/app/models/manageiq/providers/lenovo/physical_infra_manager/refresher.rb
@@ -1,8 +1,7 @@
 module ManageIQ::Providers::Lenovo
   class PhysicalInfraManager::Refresher < ManageIQ::Providers::BaseManager::Refresher
     def parse_legacy_inventory(ems)
-      log_header = "MIQ_LENOVO(#{self.class.name}.#{__method__} Calling for [#{ems.name}])"
-      $log.info(log_header)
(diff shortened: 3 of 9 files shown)


caiocmpaes:

caiocmpaes:
@miq-bot remove_label unmegeable

caiocmpaes:

caiocmpaes:

caiocmpaes:

caiocmpaes:

caiocmpaes:
@miq-bot assign @agrare 

@agrare this PR is the next on the refactoring.

.rubocop_local.yml
```diff
@@ -3,3 +3,14 @@ GlobalVars:
...
   - $log
+
+# Ignores the file at lib folder that needs to have a naming with dashes.
+AllCops:
+  Exclude:
+  - lib/manageiq-providers-lenovo.rb
+
+# Forcing Single-quote string literals except in interpolations
+Style/StringLiterals:
+  Enabled: true
```
Fryguy: For consistency we choose at the ManageIQ common level to not allow forcing this.  In fact, many of us would have preferred double-quote only, but as a community we've chosen to not to enforce either side.

.rubocop_local.yml
```diff
@@ -3,3 +3,14 @@ GlobalVars:
...
+Style/StringLiteralsInInterpolation:
+  Enabled: true
```
Fryguy: Same.

Gemfile
```diff
@@ -9,4 +9,4 @@ gemspec
 # your gem to rubygems.org.
 
 # Load Gemfile with dependencies from manageiq
-eval_gemfile(File.expand_path("spec/manageiq/Gemfile", __dir__))
+eval_gemfile(File.expand_path('spec/manageiq/Gemfile', __dir__))
```
Fryguy: I would be careful modifying files that come from the common generator.  It makes follow-up re-generation more difficult by increasing the diff that the author needs to review.

Rakefile
```diff
@@ -3,10 +3,9 @@ require 'bundler/setup'
 begin
   require 'rspec/core/rake_task'
 
-  APP_RAKEFILE = File.expand_path("../spec/manageiq/Rakefile", __FILE__)
+  APP_RAKEFILE = File.expand_path('spec/manageiq/Rakefile', __dir__)
   load 'rails/tasks/engine.rake'
   load 'rails/tasks/statistics.rake'
-rescue LoadError
```
Fryguy: You have to rescue LoadError here.  Again, please don't modify things that are generated from core.  If you want to change this across the board, this would probably be better modifying in core in the template.

app/models/manageiq/providers/lenovo/provider.rb
```diff
@@ -1,9 +1,10 @@
+# rubocop:disable Rails/HasManyOrHasOneDependent,Rails/InverseOf
```
Fryguy: This should be ignored around the specific relationship, not for the whole file.

Fryguy:
> We had no standard about string quotes until now.

We have a standard and that standard is to not bikeshed on a standard, if that makes sense. 😆  https://github.com/ManageIQ/guides/blob/b733567dfacf1434dc65400843038131ea2034d0/.rubocop_base.yml#L235-L238 was a specific decision (unfortunately, it's from over 5 years ago, before the history)

---

90% of this PR is self-imposed rubocop changes because of the addition of the string quoting to the rubocop.yml.  Another 5% is modifying the stuff that comes from the core template engine, where I think those changes should be made first, so can you make a PR to core for that?  Can you also make a separate PR with the last 5% of changes here?

.rubocop_local.yml
caiocmpaes: Ok, Jason. Good point.

app/models/manageiq/providers/lenovo/provider.rb
caiocmpaes: Right.

caiocmpaes:
Thank You for your review, @Fryguy . This observation about the 5% is totally true. The big parts of this PR were about the string quotes. I will make the changes.

caiocmpaes:
@Fryguy Do you have more requests about this PR?

app/models/manageiq/providers/lenovo/physical_infra_manager/refresher.rb
```diff
@@ -2,7 +2,7 @@ module ManageIQ::Providers::Lenovo
   class PhysicalInfraManager::Refresher < ManageIQ::Providers::BaseManager::Refresher
     def parse_legacy_inventory(ems)
       log_header = "MIQ_LENOVO(#{self.class.name}.#{__method__} Calling for [#{ems.name}])"
-      $log.info("#{log_header}")
+      $log.info(log_header)
```
agrare: Prefer to just log the string directly, log_header isn't being used anywhere else in this method so no need for a var

app/models/manageiq/providers/lenovo/manager_mixin.rb
```diff
@@ -74,7 +74,7 @@ def validate_connection(connection)
 
     def connection_rescue_block
       yield
-    rescue => err
+    rescue StandardError => err
```
agrare: I may be in the minority but I don't think `rescue StandardError => err` is better than just `rescue => err`

app/models/manageiq/providers/lenovo/provider.rb
```diff
@@ -1,9 +1,10 @@
 class ManageIQ::Providers::Lenovo::Provider < ::Provider
+  # rubocop:disable Rails/HasManyOrHasOneDependent,Rails/InverseOf
```
agrare: Please don't do this, either add the `:dependent` and `:inverse` or just leave it and its okay if rubocop complains

app/models/manageiq/providers/lenovo/provider.rb
```diff
@@ -1,9 +1,10 @@
...
   has_one :ph_infra_manager,
-          :foreign_key  =>  "provider_id",
-          :class_name   =>  "ManageIQ::Providers::Lenovo::PhysicalInfraManager",
-          :autosave     => true
-
-  validates   :name, :presence => true, :uniqueness =>  true  
+          :foreign_key => "provider_id",
+          :class_name  => "ManageIQ::Providers::Lenovo::PhysicalInfraManager",
+          :autosave    => true
+  # rubocop:enable Rails/HasManyOrHasOneDependent,Rails/InverseOf
```
agrare: Same as above

douglasgabriel:
Since @caiocmpaes doesn't work with us anymore, I'm assuming his PRs. So I took the liberty to address @agrare comments. If there is more things to address here, please let me know :+1:

PR #183, opened 2018-05-23
Moved common dictionary constants to component parser (#7)

caiocmpaes:
**What this PR does:**
- The specialized parsers inherit from `ComponentParser`. Thus, all maps that were used in more than one parser can live inside the parent class and be used directly by Its children (without namespacing).

**Reason:**
- This enables a better cohesion on the parser classes. The maps that are common to more than one parser are available in the parent class.

### DIFF
--- a/app/models/manageiq/providers/lenovo/physical_infra_manager/parser/component_parser.rb
+++ b/app/models/manageiq/providers/lenovo/physical_infra_manager/parser/component_parser.rb
@@ -4,43 +4,88 @@ module ManageIQ::Providers::Lenovo
   # to a MiQ format
   #
   class PhysicalInfraManager::Parser::ComponentParser
-    #
-    # Returns a hash containing the structure described on dictionary
-    # and with the values in the source.
-    #
-    # @param source     - Object that will be parse to a hash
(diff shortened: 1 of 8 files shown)


caiocmpaes:

caiocmpaes:

caiocmpaes:

caiocmpaes:

app/models/manageiq/providers/lenovo/physical_infra_manager/parser/component_parser.rb
```diff
@@ -4,43 +4,88 @@ module ManageIQ::Providers::Lenovo
...
   #
   class PhysicalInfraManager::Parser::ComponentParser
-    #
-    # Returns a hash containing the structure described on dictionary
-    # and with the values in the source.
-    #
-    # @param source     - Object that will be parse to a hash
-    # @param dictionary - Hash containing the instructions to translate the object into a Hash
-    #
-    # @see ParserDictionaryConstants
```
felipedf: This is more a reminder to get rid of that reference when the `ParserDictionaryConstants` is deleted

PR #202, opened 2018-06-21
Adding location led name

douglasgabriel:
Some identify LEDs of some components has the value `"Location"` in its names


douglasgabriel:
@miq-bot assign @agrare

PR #205, opened 2018-07-03
Parsing physical_swtich_id and physical_chassis_id into events

felipedf:
This PR is able to parse `physical_switch_id` into events, so we can handle events for switches.
Depends on: https://github.com/ManageIQ/manageiq-schema/pull/229

### DIFF
--- a/app/models/manageiq/providers/lenovo/physical_infra_manager/event_parser.rb
+++ b/app/models/manageiq/providers/lenovo/physical_infra_manager/event_parser.rb
@@ -1,38 +1,64 @@
 module ManageIQ::Providers::Lenovo::PhysicalInfraManager::EventParser
-  def self.event_to_hash(data, ems_id)
-    event = filter_data(data)
-    event_hash = {
-      :event_type         => event[:event_type],
-      :ems_ref            => event[:ems_ref],
-      :source             => event[:source],
-      :physical_server_id => get_physical_server_id(event[:component_id]),
-      :message            => event[:message],
-      :timestamp          => event[:timestamp],
-      :full_data          => event,
-      :ems_id             => ems_id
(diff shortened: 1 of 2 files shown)


app/models/manageiq/providers/lenovo/physical_infra_manager/event_parser.rb
```diff
@@ -32,7 +36,29 @@ def self.filter_data(data)
...
+      if event[:parent_type] == CHASSIS
+        event_resources[:physical_chassis_id] = get_klass_id(PhysicalChassis, event[:parent_uuid])
+      end
+
+      if event[:component_type] == SWITCH
+        event_resources[:physical_switch_id] = get_klass_id(PhysicalSwitch, event[:component_id])
+      elsif SERVER.include?(event[:component_type])
+        event_resources[:physical_server_id] = get_klass_id(PhysicalServer, event[:component_id])
+      else
+        raise(NotImplementedError, "The event of type #{event[:component_type]} is not supported")
```
CharlleDaniel: I think that is a little better only create a new log output like `$log.erro(message)` instead of raise an exception. @felipedf do you agree ?

app/models/manageiq/providers/lenovo/physical_infra_manager/event_parser.rb
```diff
@@ -32,7 +36,29 @@ def self.filter_data(data)
...
+      end
+
+      event_resources
+    end
+
+    def get_klass_id(klass, uid_ems)
```
douglasgabriel: __Suggestion__
In this method you are getting the id of the event resource, so I would say that the name `get_resource_id` fit better the goal of the method, don't you think?

PR #217, opened 2018-08-06
Adding and classifying error messages

matheuscmelo:
Adding error messages to settings.yml
This was possible by an automatic analysis of LXCA's documentation.

PR #218, opened 2018-08-09
config/settings - add missing fields, fixing ui-classic breakage

himdel:
Every event group in settings needs to have 3 arrays: `critical`, `detail`, `warning`.

ManageIQ/manageiq-providers-lenovo#217 introduces a couple of groups missing one or two of the required arrays, which breaks 54 ui-classic specs :): https://gist.github.com/himdel/c7b682676542936c8e4533ee58da7ebb (example https://travis-ci.org/ManageIQ/manageiq-ui-classic/jobs/413927437)

Adding the missing ones here.

Cc: @matheuscmelo, @CharlleDaniel , @agrare

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -71,6 +71,7 @@
           - 'IMM0133'
           :detail:
           - 'PLAT0187'
+          :warning: []
           :name: Firmware
         :devices:
           :critical:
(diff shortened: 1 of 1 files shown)


himdel:
@miq-bot add_label bug, gaprindashvili/no

PR #220, opened 2018-08-20
Implementing Ansible Runner engine for Lenovo's provider

douglasgabriel:
__This PR is able to__
- Add the capability to run Ansible resources (roles and playbooks) against a LXCA or its components;

__RFC__
- How to embed https://github.com/lenovo/ansible.lenovo-lxca into `manageiq-providers-lenovo`

__Usage__

Run the following rake task to fetch roles with ansible-galaxy
```bash
bundle exec rake evm:ansible_runner:seed
```

After this, the following script is possible
```ruby
ph_manager_lenovo = ManageIQ::Providers::Lenovo::PhysicalInfraManager.first
ph_manager_lenovo.run_ansible({'role_name' => 'lenovo.lxca-inventory'})
```

### DIFF
--- a/app/models/manageiq/providers/lenovo/physical_infra_manager/operations/ansible_sender.rb
+++ b/app/models/manageiq/providers/lenovo/physical_infra_manager/operations/ansible_sender.rb
@@ -1,0 +1,98 @@
+#
+# Mixin to run ansible operations
+#
+module ManageIQ::Providers::Lenovo::PhysicalInfraManager::Operations::AnsibleSender
+  extend ActiveSupport::Concern
+
+  #
+  # Executes ansible resources inside +root_dir+
+  #
+  # @param args [Hash] - The args to run the ansible resource.
+  # @option args [String] 'playbook_name' - The name of the playbook to be executed.
+  #         If this param is filled a yaml file that matches with the specified name
+  #         will be searched inside +playbooks_dir+.
+  #         In cases where the playbook is not in +playbooks_dir+ root, please, specify
+  #         the relative path to it.
+  #         Example:
+  #           run_ansible({'playbook_name' => 'config/firmware_update_all_dev'})
+  #
+  # @option args [String] 'role_name' - The name of the role to be executed.
+  #        If this param is filled it will search the role inside +roles_dir+.
+  #        Example:
+  #           run_ansible({'role_name' => 'lenovo.lxca-inventory'})
+  #
+  # @option args [Hash] vars - The set of vars needed to execute the ansible resource
+  #
+  # @return Ansible::Runner::Response
+  #
+  # @see #root_dir
+  # @see #playbooks_dir
+  # @see #roles_dir
+  #
+  def run_ansible(args = {})
+    playbook_name, role_name, vars = load_params(args)
+    if playbook_name.present?
+      Ansible::Runner.run(
+        {},
+        ansible_default_vars.merge(vars),
+        playbooks_dir.join(playbook_name)
+      )
+    else
+      Ansible::Runner.run_role(
+        {},
+        ansible_default_vars.merge(vars),
+        role_name,
+        :roles_path => roles_dir
+      )
+    end
+  end
+
+  #
+  # The set of default vars to run playbooks.
+  #
+  # @return [Hash] containing the default vars to run playbooks
+  #
+  def ansible_default_vars
+    auth = authentications.first
+
+    {
+      'lxca_user'     => auth.userid,
+      'lxca_password' => auth.password,
+      'lxca_url'      => "https://#{hostname}",
+    }
+  end
+
+  private
+
+  #
+  # The root dir where ansible lives
+  #
+  def root_dir
+    ManageIQ::Providers::Lenovo::Engine.root.join("content/ansible_runner")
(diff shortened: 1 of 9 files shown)


.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "content_tmp/ansible"]
```
agrare: @douglasgabriel this likely won't work when this is bundled from the main app (aka most developer workflows and the build system).

douglasgabriel: Yeah I had a feeling that it won't work, but, at this point, I am just doing some tests about how to run the ansible inside our provider, and I was not concern on how we will serve this content. But, anyway, is there some approach to embed this kind of content in ManageIQ?

douglasgabriel:
@miq-bot assign @agrare

.gitmodules
agrare: I wouldn't wait for a solution to this, we don't have anything planned.  For now I would copy&paste the playbooks here.  You should be able to seed the roles from github with the requirements.yml.

douglasgabriel: Hmm, got it

douglasgabriel:
For now I decided to remove the playbooks repository from this PR, once importing roles via `ansible-galaxy` is already working for us. If it is necessary in the future, we can open another PR.

@agrare could you review this?

app/models/manageiq/providers/lenovo/physical_infra_manager/operations/ansible_sender.rb
```diff
@@ -0,0 +1,98 @@
...
+  #
+  # @return Ansible::Runner::Response
+  #
+  # @see #root_dir
+  # @see #playbooks_dir
+  # @see #roles_dir
+  #
+  def run_ansible(args = {})
+    playbook_name, role_name, vars = load_params(args)
+    if playbook_name.present?
```
agrare: I really don't like methods that allow callers to throw anything at them and they figure it out from there.
Much prefer different methods for `.run` vs `.run_role`, is an abstraction layer even needed in that case?  Seems like you could just call `Ansible::Runner` directly.

app/models/manageiq/providers/lenovo/physical_infra_manager/operations/ansible_sender.rb
```diff
@@ -0,0 +1,98 @@
...
+      Ansible::Runner.run(
+        {},
+        ansible_default_vars.merge(vars),
+        playbooks_dir.join(playbook_name =~ /\.yml/ ? playbook_name : "#{playbook_name}.yml")
```
agrare: Similar point here, why not just enforce that you need the full path?  We own all of the callers so it isn't for end-user ease-of-use

app/models/manageiq/providers/lenovo/physical_infra_manager/operations/ansible_sender.rb
douglasgabriel: I see. Well the abstraction that this method brings is add default variables and playbooks directory to `Ansible::Runner` calls. Nothing more fair that the provider abstract these information, once the playbook dir lives in `manageiq-providers` project and it owns the provider vars. 

The `run` and `run_role` methods already exists on `Ansible::Runner` so this method also adds an abstraction on it, as well. This way, this method will be useful when we implement more specific methods, like `ansible_update_firmware` or `ansible_provisioning_server` for example, where they will just call the same method, changing only the variables. Beside that it is being already useful in our [Api implementation](https://github.com/ManageIQ/manageiq-api/pull/474) where it allow clients to call any ansible resource (playbook or role) in a common way, making easier to implement ansible based buttons on UI.

app/models/manageiq/providers/lenovo/physical_infra_manager/operations/ansible_sender.rb
douglasgabriel: I agree with you here, just joining `playbook_name` is enough

app/models/manageiq/providers/lenovo/physical_infra_manager/operations/ansible_sender.rb
agrare: > I see. Well the abstraction that this method brings is add default variables and playbooks directory to Ansible::Runner calls. Nothing more fair that the provider abstract these information, once the playbook dir lives in manageiq-providers project and it owns the provider vars.

I don't have any issue with the provider having some default variables

> The run and run_role methods already exists on Ansible::Runner so this method also adds an abstraction on it, as well.

This is what I have an issue with, a method that can accept any combination of arguments that cause it to do different things is harder to debug later and harder to assert the parameters are valid.

If you want to have one method to do it all that's up to you

douglasgabriel: > This is what I have an issue with, a method that can accept any combination of arguments that cause it to do different things is harder to debug later and harder to assert the parameters are valid.

I see your point, and I agree. But, since it is a decision that have to be made at some point (API call or `run_ansible` call) we decide, at least for now, leave this logic here. Thanks for your tips, I'll try to figure out a better implementation for it

PR #223, opened 2018-08-29
Add pagination on Event Catcher to fix EVM log bug

EsdrasVP:
This PR is able to:
- Add pagination on Event Catcher using header `Range` from XClarity Client to fix EVM log bug
- Supress some `info` and `debug` logs, only allowing `error` ones to decrease log size on disk
- Fix connection timeout param assignment

Depends on:
- [~lenovo/xclarity_client#126 - Merged~](https://github.com/lenovo/xclarity_client/pull/126)
- [~lenovo/xclarity_client#127 - Merged~](https://github.com/lenovo/xclarity_client/pull/127)
- [~lenovo/xclarity_client#132 - Merged~](https://github.com/lenovo/xclarity_client/pull/132)
- [~ManageIQ/manageiq-providers-lenovo#228~](https://github.com/ManageIQ/manageiq-providers-lenovo/pull/228)

Related to BZ:  https://bugzilla.redhat.com/show_bug.cgi?id=1629905

### DIFF
--- a/app/models/manageiq/providers/lenovo/physical_infra_manager/event_catcher/stream.rb
+++ b/app/models/manageiq/providers/lenovo/physical_infra_manager/event_catcher/stream.rb
@@ -1,10 +1,18 @@
 class ManageIQ::Providers::Lenovo::PhysicalInfraManager::EventCatcher::Stream
+  # The following codes represent the event classes that we don't want on our requests.
+  # Each event class corresponds to the source of the event, them being:
+  #  - 200: Audit
+  #  - 800: Rack or tower server
+  DISREGARDED_EVENTS = %w(200 800).freeze
+
+  #
   # Creates an event monitor
   #
   def initialize(ems)
+    @timeout, @events_pool_percentage = read_event_settings
     @ems = ems
     @collect_events = true
-    @last_event_ems_ref = last_event_ems_ref(@ems.id)
+    @last_event_ems_ref = last_event_ems_ref(@ems.id) || 0
   end
 
   # Stop capturing events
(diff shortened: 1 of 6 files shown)


EsdrasVP:
@miq-bot assign @douglasgabriel

app/models/manageiq/providers/lenovo/physical_infra_manager/event_catcher/stream.rb
```diff
@@ -1,10 +1,13 @@
 class ManageIQ::Providers::Lenovo::PhysicalInfraManager::EventCatcher::Stream
+  TIMEOUT = 90
```
douglasgabriel: A quick comment would be welcome here, explaining the number and its quantity (seconds, minutes, hours)

app/models/manageiq/providers/lenovo/physical_infra_manager/event_catcher/stream.rb
```diff
@@ -44,29 +47,45 @@ def parse_events(events)
...
     parsed_events
   end
 
   def events
     expression = { :filterType => 'FIELDNOTREGEXAND', :fields => filter_fields }
-
     opts = {'filterWith' => expression.to_json}
 
+    @last_cn ||= get_last_cn(opts)
+    opts['headers'] = pagination unless @last_event_ems_ref >= @last_cn
```
douglasgabriel: __Suggestion__
I think that the positive condition would be better understood. In this case:
`opts['headers'] = pagination if @last_event_ems_ref < @last_cn`

app/models/manageiq/providers/lenovo/physical_infra_manager/event_catcher/stream.rb
```diff
@@ -44,29 +47,45 @@ def parse_events(events)
...
     connection.fetch_events(opts)
   end
 
+  def get_last_cn(opts)
+    headers = { :range => 'item=0-1' }
+    opts['headers'] = headers
+    connection.get_last_cn(opts) # get last event cn based on the response
+  end
+
+  def pagination
```
douglasgabriel: Also a comment about this routine would be welcome. Mainly explaining this calculation `@limit = @last_cn * 0.07 + offset`.

If this `0.07` could be a variable with a more semantical name, the method understanding would be improved

EsdrasVP:
@douglasgabriel Just made the fixes that you requested. Is there any other thing that you want to highlight?

app/models/manageiq/providers/lenovo/physical_infra_manager/event_catcher/stream.rb
```diff
@@ -24,16 +26,11 @@ def each_batch
   private
 
   def filter_fields
```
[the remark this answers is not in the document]
EsdrasVP: @douglasgabriel You're right, I changed it so these values can be inside of a constant. I also changed the `filter_fields` assignment to avoid repetiton.

PR #235, opened 2018-10-03
Parse disks into canisters

EsdrasVP:
This PR is able to:
- Parse PhysicalDisks inside a Canister
- Generate a `ems_ref` for a PhysicalDisk, since it doesn't have any field that resembles to an unique ID

Depends on:
- [ManageIQ/manageiq-schema#285](https://github.com/ManageIQ/manageiq-schema/pull/285)
- [ManageIQ/manageiq#18053](https://github.com/ManageIQ/manageiq/pull/18053)

### DIFF
--- a/spec/models/manageiq/providers/lenovo/physical_infra_manager/physical_infra_spec_common.rb
+++ b/spec/models/manageiq/providers/lenovo/physical_infra_manager/physical_infra_spec_common.rb
@@ -1,0 +1,35 @@
+module PhysicalInfraSpecCommon
+  extend ActiveSupport::Concern
+
+  MODELS = %i(
+    ext_management_system physical_rack physical_chassis physical_server
+    physical_switch physical_storage physical_disk customization_script
+  ).freeze
+
+  def serialize_inventory
+    skip_atributes = %w(created_at updated_at last_refresh_date updated_on)
+    inventory = {}
+    PhysicalInfraSpecCommon::MODELS.each do |rel|
+      inventory[rel] = rel.to_s.classify.constantize.all.collect do |e|
+        e.attributes.except(*skip_atributes)
+      end
+    end
+
+    inventory
+  end
+
+  def assert_models_not_changed(inventory_before, inventory_after)
+    aggregate_failures do
+      PhysicalInfraSpecCommon::MODELS.each do |model|
+        expect(inventory_after[model].count).to eq(inventory_before[model].count), "#{model} count"\
+               " doesn't fit \nexpected: #{inventory_before[model].count}\ngot#{inventory_after[model].count}"
+
+        inventory_after[model].each do |item_after|
+          item_before = inventory_before[model].detect { |i| i["id"] == item_after["id"] }
+          expect(item_after).to eq(item_before), \
+                                "class: #{model.to_s.classify}\nexpected: #{item_before}\ngot: #{item_after}"
+        end
+      end
+    end
+  end
+end
--- a/app/models/manageiq/providers/lenovo/inventory/parser/component_parser/physical_disk.rb
+++ b/app/models/manageiq/providers/lenovo/inventory/parser/component_parser/physical_disk.rb
@@ -11,28 +11,32 @@ module ManageIQ::Providers::Lenovo
       :disk_size       => 'size'
     }.freeze
 
-    def lazy_build(components, storage_uuid)
+    def build(driver, index, storage = nil, canister = nil)
+      parsed_disk = parse_disk(driver, storage, index)
+
+      add_parent(parsed_disk, :belongs_to => :physical_storage, :object => storage) if storage
+      add_parent(parsed_disk, :belongs_to => :canister, :object => canister) if canister
+      physical_disk = @persister.physical_disks.build(parsed_disk)
+
+      physical_disk
+    end
+
+    def total_space(components)
       total_space = 0
-
       components.each do |component|
-        component['drives'].each do |drive|
-          properties = parse_drive(drive)
-          properties[:physical_storage] = @persister.physical_storages.lazy_find(storage_uuid)
-
-          total_space += properties[:disk_size].to_i
-
-          @persister.physical_disks.build(properties)
+        component['drives'].each do |driver|
+          total_space += driver['size'].to_i
         end
       end
-
       total_space
     end
 
-    def parse_drive(drive)
+    def parse_disk(driver, storage, index)
       result = {}
-      PHYSICAL_DISK.each_pair do |key, drive_key|
-        result[key] = drive[drive_key]
+      PHYSICAL_DISK.each_pair do |key, driver_key|
+        result[key] = driver[driver_key]
       end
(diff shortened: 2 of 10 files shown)


EsdrasVP:
@agrare Yeah I saw this earlier here and was trying to find what was wrong, it was a minor issue so I'll push the fix now.

agrare:
@EsdrasVP since we're saving physical_disks as a child collection of canisters anyway we don't need to encode the parent ems_ref in the disk's ems_ref.  Just using the disk location as long as it is always present and unique under its parent should be enough for the physical_disk.ems_ref

agrare:
> The location could be the same for two different disks in two different storages/canisters

The disk doesn't belong directly to an ems but through a canister so it is okay for two disks on different canisters to have the same location.

> It doesn't always come on the response that we parse

This is a serious issue, is there _anything_ that always comes back that uniquely identifies a disk on a canister/storage??

EsdrasVP:
> The disk doesn't belong directly to an ems but through a canister so it is okay for two disks on different canisters to have the same location.

@agrare Oh, got it, I thought that it could cause conflict on Physical Disks table.

> This is a serious issue, is there _anything_ that always comes back that uniquely identifies a disk on a canister/storage??
> 
> It sounds like neither location nor serial_number can be actually counted on to be present.

Yes, I agree, the storage endpoint that we consume lacks consistency, we only figured this problem out because we managed a storage that we didn't had before. And that's why I opted to use only its parent `ems_ref` with its index, because unlike `serial_number` or `location` its parent `ems_ref` should exist.

agrare:
@EsdrasVP but does the parent's ems_ref buy you anything?  If we don't get location or serial number back and all 16 disks just have the parent ems_ref you'll delete 15 and update 1 disk.

Ladas:
@EsdrasVP we should start wit a proper spec: ```Do a refresh, store all primary keys of all models, do second refresh and check that the records are the same```

You can use similar code as https://github.com/Ladas/manageiq-providers-azure/blob/915fda148e7978875b78c86c79c846d6654c9a7e/spec/models/manageiq/providers/azure/cloud_manager/compare_legacy_and_new_refresh_spec.rb#L20-L25

The methods are in a file https://github.com/Ladas/manageiq-providers-azure/blob/915fda148e7978875b78c86c79c846d6654c9a7e/spec/models/manageiq/providers/azure/cloud_manager/azure_refresher_spec_common.rb#L112

It's a very simple spec and it will verify the delete/recreate is not happening.

EsdrasVP:
> @EsdrasVP we should start wit a proper spec: `Do a refresh, store all primary keys of all models, do second refresh and check that the records are the same`
> 
> You can use similar code as https://github.com/Ladas/manageiq-providers-azure/blob/915fda148e7978875b78c86c79c846d6654c9a7e/spec/models/manageiq/providers/azure/cloud_manager/compare_legacy_and_new_refresh_spec.rb#L20-L25
> 
> The methods are in a file https://github.com/Ladas/manageiq-providers-azure/blob/915fda148e7978875b78c86c79c846d6654c9a7e/spec/models/manageiq/providers/azure/cloud_manager/azure_refresher_spec_common.rb#L112
> 
> It's a very simple spec and it will verify the delete/recreate is not happening.

@Ladas Correct me if I'm wrong, but this full refresh test [here](https://github.com/ManageIQ/manageiq-providers-lenovo/blob/d277834e03fac8fcba1492a25886058345941ac9/spec/models/manageiq/providers/lenovo/physical_infra_manager/refresher_spec.rb#L73) doesn't already cover that?

Ladas:
@EsdrasVP so that spec checks the count and specific disk, but it might miss if something gets deleted/recreated the second refresh (the primary key would change)

EsdrasVP:
> @EsdrasVP so that spec checks the count and specific disk, but it might miss if something gets deleted/recreated the second refresh (the primary key would change)

@Ladas I put the spec test here, could you check if everything is ok with it?

spec/models/manageiq/providers/lenovo/physical_infra_manager/physical_infra_spec_common.rb
```diff
@@ -0,0 +1,35 @@
+module PhysicalInfraSpecCommon
+  extend ActiveSupport::Concern
+
+  MODELS = %i(
+    ext_management_system physical_rack physical_chassis physical_server
+    physical_switch physical_storage
```
Ladas: does this list all models? I think e.g. physical_disk is missing

EsdrasVP: No, I forgot to put it after physical_storage, sorry about that. Now it's fixed.

agrare:
@EsdrasVP what is the status of this?  I tried applying all three of your PRs and am getting a test failure:
```
storage_physical_disk_one = PhysicalDisk.find_by(:ems_ref => "208000C0FF2683AF_0")
(byebug) storage_physical_disk_one
nil
```

EsdrasVP:
> @EsdrasVP what is the status of this? I tried applying all three of your PRs and am getting a test failure:

@agrare I ran the tests here and everything looks fine, but I think it's because I didn't rebased my branch yet to get the changes from https://github.com/ManageIQ/manageiq-providers-lenovo/pull/236. Could this be causing the error?

agrare:
@EsdrasVP still getting:

```
Failures:

  1) ManageIQ::Providers::Lenovo::PhysicalInfraManager::Refresher will perform a full refresh
     Failure/Error: expect(storage_physical_disk_one.model).to eq("ST9300653SS")
     
     NoMethodError:
       undefined method `model' for nil:NilClass
     # ./spec/models/manageiq/providers/lenovo/physical_infra_manager/refresher_spec.rb:141:in `assert_specific_storage'
     # ./spec/models/manageiq/providers/lenovo/physical_infra_manager/refresher_spec.rb:86:in `block (3 levels) in <top (required)>'
     # ./spec/models/manageiq/providers/lenovo/physical_infra_manager/refresher_spec.rb:79:in `times'
     # ./spec/models/manageiq/providers/lenovo/physical_infra_manager/refresher_spec.rb:79:in `block (2 levels) in <top (required)>'

Finished in 7.58 seconds (files took 6.15 seconds to load)
98 examples, 1 failure

Failed examples:

rspec ./spec/models/manageiq/providers/lenovo/physical_infra_manager/refresher_spec.rb:76 # ManageIQ::Providers::Lenovo::PhysicalInfraManager::Refresher will perform a full refresh

Randomized with seed 33096
```

This is merging both of your commits into master, not checking them out so if you haven't rebased in a while you should.

felipedf:
@agrare I guess it is all fixed now

agrare:
Did you test it @felipedf ?  I still have a change request on the core PR so this can't be merged yet.

felipedf:
> Did you test it @felipedf ? I still have a change request on the core PR so this can't be merged yet.

I did. Will look into the core PR then test everything again, will ping you again when I am done (:

PR #233, opened 2018-10-02
Calculate and store Physical Storage total space

EsdrasVP:
This PR is able to get disk information from `PhysicalDisks` inside a physical storage to calculate the total disk size and store it on `total_space`.

Depends on:
- [~ManageIQ/manageiq-schema#283~](https://github.com/ManageIQ/manageiq-schema/pull/283)

PR #236, opened 2018-10-05
Graph refresh

slemrmartin:
Graph refresh collects inventory to `InventoryCollection` + `InventoryObject` objects and saves by `Persister`. 

Avoids using nested hashes when collecting data. 

Switching Refresher as ManagerRefresh subclass, added collector/parser/persister classes.

Parser components are derived as much as possible from old style refresh.

---

- [x] **depends on** https://github.com/ManageIQ/manageiq/pull/18063


slemrmartin:
lenovo er

slemrmartin:
lenovo er

slemrmartin:
cc @agrare @Ladas @douglasgabriel 

There is one problem with inventories `physical_disks` and `canisters`. They didn't have unique key in old refresh and as I see they didn't have even unique key in lenovo REST API doc. 
For canisters, there should be called /canisters to get UUID.
I used `serial_number` instead, because it sounds like (yes :) unique value

slemrmartin:
cc @EsdrasVP

slemrmartin:
@EsdrasVP okay, I changed it in https://github.com/ManageIQ/manageiq/pull/18063, that unique value is storage id + serial number, which can be nil. 
Disk/canister will be badly overwritten only in case of 2 disks/canisters without serial number belonging to the same storage. 

Also there are more changes in #235, which must be integrated to this code, right? There are some disk "total space" and multi-enclosures, right?

EsdrasVP:
> Disk/canister will be badly overwritten only in case of 2 disks/canisters without serial number belonging to the same storage.

@slemrmartin Yeah, if what I did on https://github.com/ManageIQ/manageiq-providers-lenovo/pull/235 it's approved and we use `GET /canisters`, this problem should not happen.

> Also there are more changes in #235, which must be integrated to this code, right? There are some disk "total space" and multi-enclosures, right?

Yes, although the implementation of `total_space` was made on another, and already approved, PR. Don't know why is there, maybe due to a rebase, I'll check that.

slemrmartin:
@EsdrasVP all parser components are copied to new location here, because there is too many changes to use the same code for current and graph refresh. 
So during the time there will be 2 parallel refreshes, code must be written to both places.

slemrmartin:
- [x] integrated https://github.com/ManageIQ/manageiq-providers-lenovo/pull/232 to graph refresh

slemrmartin:
#233 integrated

CharlleDaniel:
@slemrmartin  I will close and reopen just for the travis run the specs again.

slemrmartin:
@CharlleDaniel this PR depends on https://github.com/ManageIQ/manageiq/pull/18063, so tests are failing now

PR #238, opened 2018-10-08
Graph refresh specs

slemrmartin:
Specs are created against old refresh, then values were checked against graph refresh

**depends on**: https://github.com/ManageIQ/manageiq-providers-lenovo/pull/236

### DIFF
--- a/spec/models/manageiq/providers/lenovo/physical_infra_manager/refresher_spec.rb
+++ b/spec/models/manageiq/providers/lenovo/physical_infra_manager/refresher_spec.rb
@@ -2,6 +2,21 @@ require_relative 'physical_infra_spec_common'
 describe ManageIQ::Providers::Lenovo::PhysicalInfraManager::Refresher do
   include PhysicalInfraSpecCommon
 
+  let(:physical_rack_ref) { "096F8C92-08D4-4A24-ABD8-FE56D482F8C4" }
+  let(:physical_server_ref) do
+    { :ems1 => "7936DD182C5311E3A8D6000AF7256738",
+      :ems2 => "BD775D06821111E189A3E41F13ED5A1A" }
+  end
+  let(:physical_storage_ref) do
+    { :ems1 => "208000C0FF2647DA",
+      :ems2 => "123400C0FF261234" }
+  end
+  let(:physical_chassis_ref) do
+    { :ems1 => "9841028B07714FD09D9297C6D1A4943E",
+      :ems2 => nil }
+  end
+  let(:physical_switch_ref) { "00000000000010008000A48CDB984B00" }
+
   before(:all) do
     vcr_path = File.dirname(described_class.name.underscore)
     options = {:allow_playback_repeats => true}
@@ -18,6 +33,19 @@ describe ManageIQ::Providers::Lenovo::PhysicalInfraManager::Refresher do
     while VCR.cassettes.last
       VCR.eject_cassette
     end
+  end
+
+  before(:each) do
+    stub_settings_merge(
+      :ems_refresh => {
+        :lenovo_ph_infra => {
+          :inventory_object_refresh => true,
+          :inventory_collections    => {
+            :saver_strategy => "default"
+          }
+        }
+      }
+    )
   end
 
   let(:auth) do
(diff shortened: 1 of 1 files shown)


slemrmartin:
Cc @Ladas @agrare @douglasgabriel 
Check only last commit (needs rebase of #236 )

slemrmartin:
integrated #232 to graph refresh
@CharlleDaniel your PR contains some bug, when I run your (current) refresh against this extended `refresher_spec`, then value of `Firmware.count` differs between 1st and 2nd round. Probably some bad unique key evaluation. 
Firmware.count in 1st round for ems1 (which isn't tested before) gives 22 Firmwares https://github.com/ManageIQ/manageiq-providers-lenovo/pull/238/files#diff-ff337ad5f24e45e4fbe0d6a7a21eed09R159

slemrmartin:
#233 integrated

Ladas:
@slemrmartin can you rebase?

slemrmartin:
@douglasgabriel Are you from Lenovo? Please notice that Refresh spec is now using graph refresh code instead old code. 

Based on flag https://github.com/ManageIQ/manageiq-providers-lenovo/pull/238/files#diff-ff337ad5f24e45e4fbe0d6a7a21eed09R39

and https://github.com/ManageIQ/manageiq-providers-lenovo/pull/236/files#diff-646ad6b10917e4385cbcc8c8524e24a2R373 (which will switch using graph refresh in production)

slemrmartin:
@EsdrasVP pls notice ^^^

spec/models/manageiq/providers/lenovo/physical_infra_manager/refresher_spec.rb
```diff
@@ -17,6 +32,19 @@
     end
   end
 
+  before(:each) do
+    stub_settings_merge(
```
Ladas: we should be testing the old refresh too, if we are keeping it around

also e.g. on Azure we test that old refresh and new refresh result in the same DB records.

slemrmartin: it's tested, I'll add it

spec/models/manageiq/providers/lenovo/physical_infra_manager/refresher_spec.rb
```diff
@@ -17,142 +32,631 @@
...
+          expect(PhysicalStorage.count).to eq(3)
+          expect(PhysicalChassis.count).to eq(1)
+          expect(PhysicalSwitch.count).to eq(1)
+          expect(AssetDetail.count).to eq(8)
+          expect(PhysicalDisk.count).to eq(12)
+          expect(Canister.count).to eq(6)
+          expect(ComputerSystem.count).to eq(10)
+          expect(Hardware.count).to eq(11)
+          expect(GuestDevice.count).to eq(14)
+          # expect(Firmware.count).to eq(20)
```
slemrmartin: @douglasgabriel This line is not true for first round of old refresh. It's equal only in 2nd round and both rounds of graph refresh

CharlleDaniel:
> @douglasgabriel Are you from Lenovo? Please notice that Refresh spec is now using graph refresh code instead old code.
> 
> Based on flag https://github.com/ManageIQ/manageiq-providers-lenovo/pull/238/files#diff-ff337ad5f24e45e4fbe0d6a7a21eed09R39
> 
> and https://github.com/ManageIQ/manageiq-providers-lenovo/pull/236/files#diff-646ad6b10917e4385cbcc8c8524e24a2R373 (which will switch using graph refresh in production too)

@slemrmartin thanks for you report, I found a little problem in the firmware parse of Guest devices and I did open a PR to fix this.

spec/models/manageiq/providers/lenovo/physical_infra_manager/refresher_spec.rb
CharlleDaniel:
```suggestion
         expect(Firmware.count).to eq(20)
```
Please rebase this PR and remove the `#`

slemrmartin: @CharlleDaniel done

slemrmartin:
@Ladas please merge

slemrmartin:
merged #235 `refresher_spec` changes

PR #245, opened 2018-10-31
Make the port uid_ems unique

skovic:
Currently, the UUID of the device associated with a port is concatenated with a port number and used as a port's _uid_ems_; however, if a device does not have a UUID, the PCI bus number concatenated with the PCI device number and the port number is used as a port's _uid_ems_, but this combination is not guaranteed to be unqiue among providers. To ensure that the _uid_ems_ is unique, we can use a port's MAC address as its _uid_ems_ since MAC addresses are unique. In the unlikely case that a port does not have a MAC address, we will fall back on using the PCI bus number concatenated with the PCI device number and the port number as a port's _uid_ems_.


skovic:
Need to add spec tests but waiting for the PR containing spec test updates for the graph refresh to be merged first

PR #247, opened 2018-11-09
Enhancement on PhysicalDisks test

felipedf:
* Add tests to verify the condition where a `physical_disk` belongs to both `physical_storage` and `canister`
* Adapt the graph refresh to address to those tests
* Add `physical_disks` and `canisters` to the base physical_infra_manager persister https://github.com/ManageIQ/manageiq/pull/18053#issuecomment-436645792


felipedf:
@agrare

PR #269, opened 2019-05-17
Fix factory inheritence to fix issues with defaults (zone)

bdunne:
Other repos using this factory aren't expecting to have to set zone and other attributes since they're handled by the `:ext_management_system` base factory.  Setting the parent factory fixes this.


bdunne:
@miq-bot assign @agrare

PR #273, opened 2019-07-09
Replace local inventory-related definitions

tadeboro:
Since all of the relationships and inventory definitions that we are using are included in the core repo, we can drop the local definitions and replace them with the definitions from the core repo.

This PR depends on:

  * [ ] https://github.com/ManageIQ/manageiq/pull/18953

@miq-bot assign @agrare

### DIFF
--- a/app/models/manageiq/providers/lenovo/inventory/persister/definitions/physical_infra_collections.rb
+++ b/app/models/manageiq/providers/lenovo/inventory/persister/definitions/physical_infra_collections.rb
@@ -2,15 +2,29 @@ module ManageIQ::Providers::Lenovo::Inventory::Persister::Definitions::PhysicalI
   extend ActiveSupport::Concern
 
   def initialize_physical_infra_inventory_collections
-    %i(canisters
-       customization_scripts
-       physical_chassis
-       physical_disks
-       physical_racks
-       physical_servers
-       physical_storages
-       physical_switches).each do |name|
-
+    %i[
+      canisters
+      customization_scripts
+      physical_chassis
+      physical_chassis_details
+      physical_chassis_computer_systems
+      physical_chassis_hardwares
+      physical_disks
+      physical_racks
+      physical_servers
+      physical_server_details
+      physical_server_computer_systems
+      physical_server_hardwares
+      physical_server_network_devices
+      physical_server_storage_adapters
+      physical_storages
+      physical_storage_details
+      physical_storage_computer_systems
+      physical_storage_hardwares
+      physical_switches
+      physical_switch_details
+      physical_switch_hardwares
+    ].each do |name|
       add_collection(physical_infra, name)
     end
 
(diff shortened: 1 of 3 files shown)


tadeboro:
The code in this PR fails on my development machine with this error:

```
ActiveRecord::HasManyThroughSourceAssociationNotFoundError:
       Could not find the source association(s) :management_devices in model Hardware. Try
       'has_many :physical_chassis_management_devices,
       :through => :physical_chassis_hardwares, :source => <name>'. Is it one of
       vm_or_template, vm, miq_template, host, computer_system, physical_switch, networks,
       firmwares, disks, hard_disks, floppies, cdroms, partitions, volumes, guest_devices,
       storage_adapters, nics, ports, physical_ports, or connected_physical_switches?
```

I guess inventory refresh is using `::Hardware` instead of `ManageIQ::Providers::Lenovo::PhysicalInfraManager::Hardware` somewhere, but I have no idea where or why since I left the hardware-related definitions intact. But I am clearly doing something wrong.

agrare:
> I guess inventory refresh is using ::Hardware instead of ManageIQ::Providers::Lenovo::PhysicalInfraManager::Hardware somewhere, but I have no idea where or why since I left the hardware-related definitions intact. But I am clearly doing something wrong.

Hm @slemrmartin can you help here?

slemrmartin:
hey @tadeboro I'm afraid it doesn't work this way. These relationships are not included in core repo (at least I don't see where). You're replacing polymorphic associations with N:M, like `physical_chassis_computer_systems` (or `physical_storage_computer_systems` etc.). For example if you removed this specific definition, then it'll expect you have `PhysicalChassisComputerSystem` model. But what these associations says is, that you have 3 different associations to `computer_systems` (with different parent `PhysicalChassis`, `PhysicalStorage` and `PhysicalServer`. It can't be just removed. 
See attached diagram

slemrmartin:
Lenovo ER

slemrmartin:
it means you have to have unique association chain from `ExtManagementSystem` to `ComputerSystem` (and `Hardware`) for each possible diagram path

tadeboro:
I did not remove those `has_many` relationships, I just moved them to the core repo: https://github.com/ManageIQ/manageiq/pull/18953/files#diff-54a07cbbdbeffee51b848dc93f209776

And the error I posted is suggesting that the `Hardware` model is missing a `has_many` relationship, which is true since this relationship is only present in the Lenovo-specific `Hardware` model.

As I said before, somehow moving the `has_many` definitions to the parent class made the refresh process start using the wrong `Hardware` class.

tadeboro:
Here is the complete rspec output: https://gist.github.com/tadeboro/74e9c5392d22e2b12490d0efc40560bc

slemrmartin:
hm, I didn't implement it in core's definitions because it's Lenovo specific. Does it work for Redfish?
Anyway, I'll look at that rspec. As I remember management_devices were more difficult.

slemrmartin:
I think that it's some bug in associations. `has_many` association defined in parent EMS instantiates hardware badly even if it should have correct class (I don't know why).
Try to move this association back to lenovo's provider if it solves it. 
I'm not sure if it was this case but it definitely happened to me before.

tadeboro:
As discussed on Gitter, I moved the `management_devices` relationship into the `::Hardware` model removed the Lenovo-specific subclass. After this change, tests are green on my development machine.

PR #280, opened 2019-10-10
Remove to_json from toolbar data as it's not needed

ZitaNemeckova:
After https://github.com/ManageIQ/manageiq-ui-classic/pull/5997 `to_json` is not needed and will break the data format.

Related to https://github.com/ManageIQ/manageiq-ui-classic/pull/6289 fix but **not** dependent either way.

Before:
Buttons `Firmware Update` and `Provision ` don't work on click.
After:
Buttons work as expected.

@miq-bot add_label ivanchuk/no


agrare:
:+1: thanks @ZitaNemeckova !

PR #287, opened 2019-11-12
Add params_for_create and verify_credentials

agrare:
https://github.com/ManageIQ/manageiq/issues/18818


app/models/manageiq/providers/lenovo/manager_mixin.rb
```diff
@@ -44,11 +44,80 @@ def console_supported?
...
+            :type      => "number",
+          },
+          {
+            :component => "checkbox",
+            :name      => "endpoints.default.verify_ssl",
+            :label     => "Verify SSL"
+          },
+          {
+            :component => "text-field",
+            :name      => "endpoints.default.user_agent_label",
```
agrare: Not sure what user_agent_label is but the UI sends false to raw_connect

PR #291, opened 2020-03-05
Remove legacy refresh parser

d-m-u:
cause of the move to graph refresh this should all be able to be cut

It's quite likely there's stuff I missed but it's also quite likely I need someone else's eyes on this to check that.


d-m-u:
@miq-bot remove_label wip
@miq-bot add_label technical debt

app/models/manageiq/providers/lenovo/physical_infra_manager/refresher.rb
```diff
@@ -1,9 +0,0 @@
-module ManageIQ::Providers::Lenovo
-  class PhysicalInfraManager::Refresher < ManageIQ::Providers::BaseManager::Refresher
```
agrare: @d-m-u I think we should keep the `Lenovo::PhysicalInfraManager::Refresher` class even if it is empty

d-m-u: cool, just a second

PR #323, opened 2020-11-18
[KASPAROV] yarn: lock versions by adding yarn.lock

himdel:
parent https://github.com/ManageIQ/manageiq-ui-classic/pull/7515


himdel:
@miq-bot assign simaishi
@miq-bot add_label dependencies

PR #343, opened 2021-12-10
converting supported_for_create to supports :create

kbrock:
part of ManageIQ/manageiq#21573


app/models/manageiq/providers/lenovo/provider.rb
```diff
@@ -5,4 +5,6 @@ class ManageIQ::Providers::Lenovo::Provider < ::Provider
           :autosave    => true
 
   validates :name, :presence => true, :uniqueness => true
+
+  supports :create
```
agrare, line 9: Do we check the `Provider` class for `supports?(:create)` anywhere?

PR #373, opened 2022-05-10
Remove urijs dependency by using strings for API calls

Fryguy:
None of the other providers use urijs instead opting for normal template
strings. urijs is giving problems during build where it hangs on s390x
builds. In addition there have been a number of security issues with
urijs, and this is the only location that uses it.

@kavyanekkalapu @jeffibm I'm not sure how to test this...can you help me out?

cc @agrare

### DIFF
--- a/app/javascript/components/apply-config-pattern-form-provider.jsx
+++ b/app/javascript/components/apply-config-pattern-form-provider.jsx
@@ -1,7 +1,6 @@
 import React from "react";
 import PropTypes from "prop-types";
 import {connect} from "react-redux";
-import URI from "urijs";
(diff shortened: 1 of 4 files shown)


agrare:
@kavyanekkalapu these are used in the Lenovo provider under Provision:
image

agrare:
The apply-config-pattern dialog looks a little wonky, it wants you to scroll when you go to select a server and cuts off the dropdown
image
But I don't believe this to be related to this PR

agrare:
When you select a physical server, looks good
image

I do get some notification errors though
image

PR #449, opened 2026-04-01
Convert react bootstrap components to carbon react

GilbertCherrie:
Convert the React bootstrap components to Carbon React components.

Before:

Apply Config Pattern form
Screenshot 2026-04-01 at 11 08 12 AM

Firmware Update form
Screenshot 2026-04-01 at 11 08 17 AM

After:

Apply Config Pattern form
Screenshot 2026-04-01 at 11 10 28 AM

Firmware Update form
Screenshot 2026-04-01 at 11 10 53 AM


GilbertCherrie:
@miq-bot assign @Fryguy 
@miq-bot add-label enhancement

Fryguy:
Not sure how difficult or not this is, but can we do the rename in one commit and any code changes in a separate commit - it's nearly impossible to review when it's delete and change in the same commit.

GilbertCherrie:
@Fryguy I removed the rename changes, I will add those back in a final commit / follow up pr assuming everything else here is good

GilbertCherrie:
@miq-bot remove-label wip

Fryguy:
@asirvadAbrahamVarghese Please review. I can't assign for some reason.

app/javascript/components/form/fields/config_pattern_field.js
```diff
@@ -18,62 +18,63 @@ class ConfigPatternField extends React.Component {
 
   getValidationState() {
```
asirvadAbrahamVarghese: I think this method is not used, validation logic is already available in the render method below:
`invalid={this.state.touched && !this.state.valid}`

app/javascript/components/form/fields/physical_server_field.js
```diff
@@ -19,60 +20,57 @@ class PhysicalServerField extends React.Component {
...
-            </FormControl>
-          </div>
-      </FormGroup>
+      <div onClick={this.onClick}>
+        <MultiSelect
+          id="selectServer"
+          titleText={__('Physical Server')}
+          label={__('Choose a Server')}
+          items={serverItems}
+          itemToString={(item) => (item ? item.label : '')}
```
asirvadAbrahamVarghese: I think even if `itemToString` is not provided it looks for a `label` property on the item object and uses that automatically

app/javascript/components/style/firmware-update-field.scss
asirvadAbrahamVarghese, line 1: Can we switch this from CSS to SCSS?

app/javascript/components/style/apply-config-pattern-form.scss
asirvadAbrahamVarghese, line 1: Same as above could this also be SCSS?

app/javascript/components/style/apply-config-pattern-form.scss
```diff
@@ -0,0 +1,7 @@
+#selectPattern {
+    margin-bottom: 20px;
+}
+
+.apply-config-pattern-form {
+    height: 300px;
+}
```
asirvadAbrahamVarghese, line 7: Do we need this margin and height? I think it looks better without them:
Image

but are we seeing any overlap issues in any resolutions when these dropdowns contain options?

app/javascript/components/apply-config-pattern-form-provider.jsx
```diff
@@ -1,25 +1,25 @@
...
-    {
-      href: href,
-      pattern_id: values["configPatternField"]
-    }));
+const applyPattern = (values) => {
+  const resources = values.physicalServerField.map(href => ({
+    href: href,
+    pattern_id: values.configPatternField
+  }));
   API.post("/api/physical_servers/", {
```
asirvadAbrahamVarghese, line 16: Just curious - what should happen if this API fails? Do we need to process the response and show an error in the UI?

app/javascript/components/apply-config-pattern-form-provider.jsx
```diff
@@ -1,25 +1,25 @@
...
     action: "apply_config_pattern_ansible",
     resources: resources,
-   });
+  });
 };
 
-const getPhysicalServerData = (providerID, patternID) => {
+const getPhysicalServerData = (providerID) => {
   const uri = `/api/physical_servers?attributes=name,href&expand=resources&filter[]=ems_id=${providerID}`;
   return API.get(uri).then((data) => data.resources.map(resource => ({
```
asirvadAbrahamVarghese, line 24: Similar to above, do we need an error handler (`.catch`) for these API calls?

app/javascript/components/apply-config-pattern-form-provider.jsx
```diff
@@ -28,86 +28,120 @@ const getPhysicalServerData = (providerID, patternID) => {
...
 
-  updateFieldsDataLoading = (name, status) => {
-    const fieldsDataLoading= this.state.fieldsDataLoading;
-    fieldsDataLoading[name]=status;
-    return fieldsDataLoading;
-  };
+    // Check if all required fields are valid
+    const newIsValid = fieldName === 'configPatternField' 
+      ? isFieldValid && values.physicalServerField && values.physicalServerField.length > 0
+      : values.configPatternField && isFieldValid;
```
asirvadAbrahamVarghese: Might be better to check `isFieldValid` once and use optional chaining instead of individually checking object keys:
```suggestion
    const newIsValid =
      isFieldValid &&
      (fieldName === "configPatternField"
        ? values?.physicalServerField?.length
        : values?.configPatternField);
```

app/javascript/components/apply-config-pattern-form-provider.jsx
```diff
@@ -28,86 +28,120 @@ const getPhysicalServerData = (providerID, patternID) => {
...
+    
+    setIsValid(newIsValid);
+  }, [values]);
 
-  componentDidMount() {
-    this.props.dispatch({
+  useEffect(() => {
+    // Initialize form buttons
+    dispatch({
```
asirvadAbrahamVarghese, line 83: Feels like we should start using `useDispatch` and `useSelector` in React components instead of the older `connect()`pattern (`mapStateToProps`, `mapDispatchToProps`).
It’s a big refactor though, so we can do that later as a follow-up since everything currently uses `connect()`.

app/javascript/components/apply-config-pattern-form-provider.jsx
```diff
@@ -28,86 +28,120 @@ const getPhysicalServerData = (providerID, patternID) => {
...
-
-  updateServers = (resourceID) => {
-    getPhysicalServerData(ManageIQ.record.recordId, resourceID)
-      .then(physicalServerList => this.setState({ physicalServerList: physicalServerList, physicalServerFieldDisabled: false }));
-  };
+  useEffect(() => {
+    // Update form button state
+    dispatch({
+      type: "FormButtons.saveable",
+      payload: isValid,
```
asirvadAbrahamVarghese, line 110: Image

I think the `Apply` button should remain disabled when no values are selected, since there’s no action on click. The current approach  using `saveable` should work fine, but can you update the [MiqButton](https://github.com/ManageIQ/manageiq-ui-classic/blob/master/app/javascript/forms/miq-button.jsx) component (which is used in the [FormButtons](https://github.com/ManageIQ/manageiq-ui-classic/blob/master/app/javascript/forms/form-buttons.jsx) component that connects to the reducer) to use the button’s `disabled` prop instead of relying on a classname to get this?

We don't need `disabled` here:
```
  const klass = ClassNames({
    btn: true,
    'btn-xs': props.xs,
    'btn-primary': props.primary,
    'btn-default': !props.primary,
  });
```
And then update HTML part to use `disabled` prop 
```
  if (btnType === 'deleteModal') {
    return (
      <Button
        className={title}
        onClick={buttonClicked}
        onKeyDown={buttonClicked}
        title={title}
        disabled={!props.enabled}
      >
        {__(props.name)}
      </Button>
    );
  }
  return (
    <Button
      className={klass}
      onClick={buttonClicked}
      onKeyDown={buttonClicked}
      title={title}
      disabled={!props.enabled}
    >
      {__(props.name)}
    </Button>
  );
```

app/javascript/components/firmware-update-form-provider.jsx
asirvadAbrahamVarghese: To be safe can you do:
```suggestion
    firmwares: resource.hardware.firmwares,
    firmwares: resource?.hardware?.firmwares || [],
```

app/javascript/components/firmware-update-form-provider.jsx
```diff
@@ -31,65 +29,72 @@ const getPhysicalServerData = (providerID) => {
...
+        setPhysicalServerList(serverList);
+        setIsLoading(false);
+      });
+  }, [dispatch, values]);
+
+  useEffect(() => {
+    // Update form button state
+    dispatch({
+      type: "FormButtons.saveable",
+      payload: isValid,
```
asirvadAbrahamVarghese, line 75: Like mentioned in `ApplyConfigPatternFormProvider` [above](https://github.com/ManageIQ/manageiq-providers-lenovo/pull/449/changes#r3092461360), I feel it’s better for the submit button to stay disabled when there are no selections

Image

app/javascript/components/style/firmware-update-field.scss
```diff
@@ -1,27 +1,95 @@
...
+
+.firmware-server-list {
+  display: flex;
+  flex-direction: column;
+  flex: 1;
+  overflow-y: auto;
+  overflow-x: hidden;
+}
+
+.firmware-server-tile {
```
asirvadAbrahamVarghese, line 36: Can you also add `display: flex` here so the server names are center aligned vertically instead of closer to the top?

Image

app/javascript/components/style/firmware-update-field.css
```diff
@@ -1,27 +1,95 @@
...
+.firmware-server-tile-content {
+  display: flex;
+  align-items: center;
+  pointer-events: none;
+  min-width: 0;
+}
+
+.firmware-server-name {
+  margin-left: 5px;
+  margin-right: 25px;
```
asirvadAbrahamVarghese: Better to convert these values to `rem`

app/javascript/components/style/firmware-update-field.scss
```diff
@@ -1,27 +1,95 @@
...
 }
 
-.firmwaresCol {
-  overflow: auto;
-  margin-right: 0;
-  height: 250px;
+.firmware-servers-column {
+  border-right: 1px solid #e0e0e0;
+  padding-right: 0 !important;
+  height: 400px;
```
asirvadAbrahamVarghese: Should this also be `rem`?

app/javascript/components/style/firmware-update-field.css
```diff
@@ -1,27 +1,95 @@
-.tabRow {
-  padding: 0px;
-  margin-left: 0px;
-  margin-right: 0px;
+.firmware-field-container {
+  margin-top: 1rem;
+  height: 400px;
```
asirvadAbrahamVarghese: I don’t think this height is needed since `firmware-servers-column` and `firmware-details-content` already have height set.
Or we could keep it and set the child nodes to `height: 100%`

app/javascript/components/style/firmware-update-field.scss
```diff
@@ -1,27 +1,95 @@
...
+  flex-shrink: 0;
+}
+
+.firmware-details-column {
+  padding-left: 2rem !important;
+  min-width: 0;
+}
+
+.firmware-details-content {
+  height: 400px;
```
asirvadAbrahamVarghese: Same as above, better if its rem

app/javascript/components/form/fields/firmware_check_list_field.js
```diff
@@ -1,48 +1,36 @@
...
+
+const FirmwareCheckListField = ({ firmwareData, parentName, updateNavItem, serverID, selectedFirmwares }) => {
+  const handleChange = (event, firmwareName) => {
+    updateNavItem(serverID, firmwareName, event.target.checked);
+  };
+
+  return (
+    <FormGroup legendText={__('Firmwares')}>
+      {firmwareData.map((firmware) => {
+        const isChecked = selectedFirmwares && selectedFirmwares[firmware.name] === true;
```
asirvadAbrahamVarghese: This can be:
```suggestion
        const isChecked = selectedFirmwares?.[firmware?.name] || false;
```
returns false if any part is falsy

app/javascript/components/form/fields/firmware_field.js
```diff
@@ -1,14 +1,8 @@
...
-  Row,
-  Tab,
-} from "react-bootstrap";
-import FirmwareCheckListField from "./firmware_check_list_field";
+import { Grid, Column, ClickableTile } from "@carbon/react";
+import { BareMetalServer, Checkmark } from "@carbon/react/icons";
+import FirmwareCheckListField from "./firmware_check_list_field.js";
 import "../../style/firmware-update-field.css";
 
 class FirmwareField extends React.Component {
```
asirvadAbrahamVarghese: Since we’re refactoring anyway, can we switch this also to a functional component if possible?

Same thought for `ConfigPatternField` and `PhysicalServerField` as well, they’re currently class components

app/javascript/components/form/fields/firmware_field.js
```diff
@@ -18,6 +12,7 @@ class FirmwareField extends React.Component {
 
     this.state = {
       navItemSelected: {},
+      selectedServerId: props.physicalServerData[0]?.id || null,
```
asirvadAbrahamVarghese: To be safe, this can be:

```suggestion
      selectedServerId: props.physicalServerData?.[0]?.id || null,
```

app/javascript/components/form/fields/firmware_field.js
```diff
@@ -27,70 +22,97 @@ class FirmwareField extends React.Component {
...
     navItemSelected[id] = values;
-    this.setState({navItemSelected})
+    this.setState({navItemSelected}, () => {
+      // Notify parent component of changes if onChange prop is provided
+      if (this.props.onChange) {
+        // Check if any firmware is selected
+        const hasSelection = Object.keys(navItemSelected).some(serverId => {
+          const serverFirmwares = navItemSelected[serverId];
+          return Object.values(serverFirmwares).some(isChecked => isChecked);
+        });
```
asirvadAbrahamVarghese: I’m thinking we don’t need this iteration here, we can use the `firmwareField` object itself after the iteration below is done to check for any selections:
```
const hasSelection = Object.keys(firmwareField).length > 0;
this.props.onChange(firmwareField, hasSelection);
```

app/javascript/components/form/fields/firmware_field.js
```diff
@@ -27,70 +22,97 @@ class FirmwareField extends React.Component {
...
-          <h4>{physicalServer.name}</h4>
-          <FirmwareCheckListField
-            updateNavItem={this.updateNavItem}
-            firmwareData={physicalServer.firmwares}
-            parentName={this.props.name}
-            serverID={physicalServer.id}/>
-        </Tab.Pane>
-      )
-    });
+    if (!physicalServerData || physicalServerData.length === 0) {
```
asirvadAbrahamVarghese: This should catch anything falsy:
```suggestion
    if (!physicalServerData?.length) {
```

app/javascript/components/form/fields/firmware_field.js
```diff
@@ -27,70 +22,97 @@ class FirmwareField extends React.Component {
...
+                  </ClickableTile>
+                );
+              })}
+            </div>
+          </Column>
+          <Column lg={10} md={4} sm={4} className="firmware-details-column">
+            {selectedServer && (
+              <div className="firmware-details-content">
+                <h4 className="firmware-section-title">{selectedServer.name}</h4>
+                <h5 className="firmware-subsection-title">{__('Firmwares')}</h5>
```
asirvadAbrahamVarghese: Do we need this `h5`? The legend below already has “Firmwares”.
Image

app/javascript/components/apply-config-pattern-form-provider.jsx
GilbertCherrie: All of our error handling is usually done by the error modal. We don't use .catch blocks and instead let the error modal display any errors that are encountered with the API response. Example: 
Screenshot 2026-04-20 at 11 27 35 AM

app/javascript/components/apply-config-pattern-form-provider.jsx
GilbertCherrie: All of our error handling is usually done by the error modal. We don't use .catch blocks and instead let the error modal display any errors that are encountered with the API response. Example: 
Screenshot 2026-04-20 at 11 27 35 AM

app/javascript/components/style/apply-config-pattern-form.scss
GilbertCherrie: Deleting the height makes the second drop down hard to see:
Screenshot 2026-04-20 at 11 22 15 AM

Keeping the height makes it easier to see the options:
Screenshot 2026-04-20 at 11 22 59 AM

app/javascript/components/apply-config-pattern-form-provider.jsx
GilbertCherrie: Yeah this should be a follow up I think

app/javascript/components/style/firmware-update-field.scss
GilbertCherrie: Where exactly did you want me to add `display: flex`? I tried it locally and didn't notice any change.

app/javascript/components/style/firmware-update-field.scss
GilbertCherrie: This value should probably stay in px. I feel like its easier to style with px then rem. If you had a specific rem value in mind I can test it out.

app/javascript/components/form/fields/firmware_field.js
GilbertCherrie: This code doesn't work, throws an error when I select a firmware value

app/javascript/components/style/firmware-update-field.scss
GilbertCherrie: This value should probably stay in px. I feel like its easier to style with px then rem. If you had a specific rem value in mind I can test it out.

app/javascript/components/form/fields/physical_server_field.js
GilbertCherrie: I'm not sure what you are referring to with this comment. Did you want to delete the itemToString prop?

app/javascript/components/form/fields/firmware_field.js
GilbertCherrie: Fixed

app/javascript/components/apply-config-pattern-form-provider.jsx
GilbertCherrie: Fixed in this PR: https://github.com/ManageIQ/manageiq-ui-classic/pull/9992. I didn't want to delete the disabled classname in this PR case it is used in other places.

app/javascript/components/firmware-update-form-provider.jsx
GilbertCherrie: Fixed in this PR: https://github.com/ManageIQ/manageiq-ui-classic/pull/9992. I didn't want to delete the disabled classname in this PR case it is used in other places.

app/javascript/components/style/firmware-update-field.scss
asirvadAbrahamVarghese: Add it for `firmware-server-tile`

**Before was aligned closer to the top:**
image

**After:**
image

app/javascript/components/form/fields/firmware_field.js
asirvadAbrahamVarghese: Can you share what’s wrong?

Note: this should be added after the `firmwareField` object is updated.
```
      if (onChange) {        
        // Convert navItemSelected to the format expected by the parent
        const firmwareField = {};
        Object.keys(updatedNavItemSelected).forEach(serverId => {
          const selectedFirmwares = [];
          Object.keys(updatedNavItemSelected[serverId]).forEach(firmwareName => {
            if (updatedNavItemSelected[serverId][firmwareName]) {
              selectedFirmwares.push(firmwareName);
            }
          });
          if (selectedFirmwares.length > 0) {
            firmwareField[serverId] = selectedFirmwares;
          }
        });
        const hasSelection = Object.keys(firmwareField).length > 0;
        onChange(firmwareField, hasSelection);
```

app/javascript/components/style/firmware-update-field.scss
asirvadAbrahamVarghese: You could use 25rem (400/16), should be equivalent(use px for smaller values like borders)

app/javascript/components/style/firmware-update-field.scss
asirvadAbrahamVarghese: Same as [above](https://github.com/ManageIQ/manageiq-providers-lenovo/pull/449#discussion_r3116962222)

app/javascript/components/form/fields/physical_server_field.js
asirvadAbrahamVarghese: Right, no need for `itemToString` if we’re picking the label from the object, it already does that

app/javascript/components/apply-config-pattern-form-provider.jsx
```diff
@@ -28,86 +28,120 @@ const getPhysicalServerData = (providerID, patternID) => {
 };
 
 const getConfigPatternData = (providerID) => {
-  const uri = `/api/customization_scripts?attributes=manager_ref,name&expand=resources&filter[]=type='ManageIQ::Providers::Lenovo::PhysicalInfraManager::ConfigPattern'&filter[]=manager_id=${providerID}`;
+  const typeFilter = encodeURIComponent("type='ManageIQ::Providers::Lenovo::PhysicalInfraManager::ConfigPattern'");
```
asirvadAbrahamVarghese: Seems like this variable is unused

app/javascript/components/form/fields/config_pattern_field.jsx
```diff
@@ -0,0 +1,61 @@
...
+ConfigPatternField.propTypes = {
+  configPatternData: PropTypes.array.isRequired,
+  name: PropTypes.string.isRequired,
+  onChange: PropTypes.func,
+  updateChildren: PropTypes.func,
+};
+
+export default ConfigPatternField;
+
+// Made with Bob
```
asirvadAbrahamVarghese: Also remove these bot-added comment(FirmwareField & PhysicalServerField have it as well)

app/javascript/components/style/apply-config-pattern-form.scss
asirvadAbrahamVarghese, line 7: For these, can you set the margin to `1.2rem` and the height to `19rem`?

GilbertCherrie:
@asirvadAbrahamVarghese fixed the issues if you can review again

asirvadAbrahamVarghese (approved):
This looks great, thanks!

I don't have merge rights. @Fryguy / @agrare, can you review/merge this?

agrare (approved):

